PR #247, opened 2025-02-23
[feat]: Add one kv store example using gcei

ByteYue:
## PR Description
Issue Number: closes #xxx

In this KV store example, users can interact with the system through three HTTP endpoints, assuming the address is 127.0.0.1:9006. First, users can submit a transaction to the system via the add_txn endpoint. If successful, the corresponding transaction hash will be returned.
``` shell
curl -X POST -H "Content-Type: application/json" -d '{
  "unsigned": {
    "nonce": 1,
    "kind": {
      "Transfer": {
        "receiver": "0x1234567890abcdef1234567890abcdef12345678",
        "amount": 100
      }
    }
  },
  "signature": "your_signature_here"
}' http://127.0.0.1:9006/add_txn

```

Subsequently, the user can query the corresponding receipt using the transaction hash.

``` shell
curl -X POST -H "Content-Type: application/json" -d '{
  "transaction_hash": "your_transaction_hash_here"
}' http://127.0.0.1:9006/get_receipt

```

Additionally, users can set a value for a specific key under an account namespace by submitting a request. Once successfully set, the value can be retrieved via the get_value endpoint.

``` shell
curl -X POST -H "Content-Type: application/json" -d '[
  "$(openssl rand -hex 20)",
  "key"
]' http://127.0.0.1:9006/get_value

```

## Tested?

- [ ] Yes
- [ ] No


bin/kvstore_v2/src/main.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    async fn run(self) {
+        loop {
+            tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+        }
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
```
Stumble, line 32: Add comments of known limitations here, i.e. TODOs for using as an example of production.

bin/kvstore_v2/src/main.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_blockchain_processing() -> Result<(), Box<dyn Error>> {
+        let _handle = Logger::try_with_str("info")
+            .unwrap()
+            .log_to_file(
+                FileSpec::default().directory("/home/jingyue/projects/gravity-sdk/kvstore_v2/logs"),
+            )
```
Stumble, line 81: hardcoded path

bin/kvstore_v2/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "pipeline_blockchain"
```
Stumble, line 2: rename? gravity_sdk_kvstore

bin/kvstore_v2/src/main.rs
ByteYue: Whoa! This is some garbage test code i forget to delete. Gonna remove them as soon as possible.

Stumble (approved):

PR #290, opened 2025-04-01
[Refactor] Refactor GCEI Protocol and Recovery API

Richard1048576:
## PR Description
Issue Number: closes #xxx

## Tested?

- [ ] Yes
- [ ] No


aptos-core/consensus/src/gravity_state_computer.rs
```diff
@@ -89,19 +89,28 @@ impl BlockExecutorTrait for GravityBlockExecutor {
...
-                            call.call(*block_id).await.unwrap();
+                get_block_buffer_manager()
+                    .set_commit_blocks(block_ids.into_iter()
+                    .enumerate()
+                    .map(|(i, x)| 
+                        {
+                            let mut v = [0u8; 32];
+                            v.copy_from_slice(block_hash.as_ref());
+                            if x == block_id {
+                                BlockIdNumHash { block_id: BlockId::from_bytes(x.as_slice()), num: block_num + (i - len + 1) as u64, hash: Some(v) }
```
Richard1048576, line 105: I think we can replace the struct name with BlockRef? Actually I think BlockMetadata or BlockInfo is better but it already used.

aptos-core/mempool/src/shared_mempool/runtime.rs
```diff
@@ -83,11 +84,10 @@ pub(crate) fn start_shared_mempool<ConfigProvider>(
 
 async fn retrieve_from_execution_routine(
     mempool: Arc<Mutex<CoreMempool>>,
-    execution_api: Arc<dyn ExecutionChannel>,
 ) {
     info!("start retrieve_from_execution_routine");
     loop {
-        match execution_api.send_pending_txns().await {
+        match get_block_buffer_manager().pop_txns(usize::MAX).await {
```
Richard1048576, line 90: Maybe we can use a variable instead of usize::MAX for it's too big.

aptos-core/mempool/src/tests/core_mempool_test.rs
```diff
@@ -860,64 +860,65 @@ fn test_parking_lot_evict_only_for_ready_txn_insertion() {
...
-        sender_bucket,
-        &vec![0].into(),
-        10,
-        None,
-        BroadcastPeerPriority::Primary,
-    );
-    assert_eq!(timeline.len(), 4);
+fn test_gc_ready_transaction() {
+    // let mut pool = setup_mempool().0;
+    // send_user_txn(&mut pool, TestTransaction::new(1, 0, 1)).unwrap();
```
Richard1048576, line 866: Will the test case be fixed later?

bin/gravity_node/src/reth_cli.rs
```diff
@@ -211,6 +187,63 @@ impl RethCli {
...
+            let block_id = self.pipe_api.get_block_id(block_ids.last().unwrap().num).unwrap_or_else(|| {
+                panic!("commit num {} not found block id", start_commit_num);
+            });
+            assert_eq!(ExternalBlockId::from_bytes(block_id.as_slice()), block_ids.last().unwrap().block_id);
+            start_commit_num = block_ids.last().unwrap().num + 1;
+            for block_id_num_hash in block_ids {
+                self.send_committed_block_info(block_id_num_hash.block_id, block_id_num_hash.hash.map(|x| B256::from_slice(x.as_slice()))).await.unwrap();
+            }
+            let last_block_number = self.provider.last_block_number().unwrap();
+            get_block_buffer_manager().remove_commited_blocks(last_block_number).await.unwrap();
```
Richard1048576, line 243: I think remove_commited_blocks can be a private func and we just set the commit block number to block_buffer_manager.

book/docs/gcei_protocol.md
```diff
@@ -111,35 +111,11 @@ There are three principles when recovery:
 3. **Find the Corresponding Recovery Root**: Match the Execution Layer’s latest block number with the closest block in
    the Consensus Layer, treating that block as the root for replay.
 4. **Compare Root with QC**: Identify blocks whose round is greater than the root’s round.
-5. **Replay Missing Blocks**: Invoke recover_ordered_block to fill in any missing blocks on the Execution Layer side.
+5. **Replay Missing Blocks**: Though buffer_manager to fill in any missing blocks on the Execution Layer side.
 
 Once the local block replay is completed, the node can communicate with peers. If the node’s round is behind during
 these communications, it initiates Block Sync to obtain the latest blocks from other nodes.
 
-### 3.4 Recovery API Definition
```
Richard1048576, line 119: We should refactor this doc before merge to main.

crates/block-buffer-manager/src/block_buffer_manager.rs
```diff
@@ -0,0 +1,393 @@
...
+    pub async fn get_ordered_blocks(
+        &self,
+        start_num: u64,
+        max_size: Option<usize>,
+    ) -> Result<Vec<(ExternalBlock, BlockId)>, anyhow::Error> {
+        if !self.is_ready() {
+            panic!("Buffer is not ready");
+        }
+        let start = Instant::now();
+        let timeout = Duration::from_secs(5);
```
Richard1048576, line 160: It should be a config of "5"

crates/block-buffer-manager/src/block_buffer_manager.rs
```diff
@@ -0,0 +1,393 @@
+use log::info;
```
Richard1048576, line 1: Because BlockBufferManger is a pure memory operation, we generally do not panic unless we encounter an irreversible error.

crates/block-buffer-manager/src/block_buffer_manager.rs
```diff
@@ -0,0 +1,393 @@
...
+    pub async fn get_commited_blocks(
+        &self,
+        start_num: u64,
+        max_size: Option<usize>,
+    ) -> Result<Vec<BlockIdNumHash>, anyhow::Error> {
+        if !self.is_ready() {
+            panic!("Buffer is not ready");
+        }
+        let start = Instant::now();
+        let timeout = Duration::from_secs(2);
```
Richard1048576, line 327: It also should be a config.

aptos-core/mempool/src/shared_mempool/runtime.rs
keanji-x: it means there is no limit, it just return the actual txns in 100ms. So i'think its ok

PR #456, opened 2025-12-04
feat: add epoch management to BlockBufferManager and RethCli

Lchangliang:
- Add current_epoch field to BlockStateMachine
- Add initial_epoch parameter to BlockBufferManager::init()
- Add epoch filtering in set_ordered_blocks() to reject old/future epoch blocks
- Add expected_epoch parameter to get_ordered_blocks() with early epoch check
- Update RethCli to use AtomicU64 for current_epoch
- Improve consume_epoch_change() to return new epoch
- Add get_current_epoch() with ready state check
- Simplify epoch filtering logic in RethCli::start_execution()
- Consolidate wait-retry logic with detailed logging

This ensures only blocks from the current epoch are processed and prevents epoch-related issues during epoch changes.

## PR Description
Issue Number: closes #xxx

## Tested?

- [ ] Yes
- [ ] No

### DIFF
--- a/aptos-core/consensus/src/block_storage/block_store.rs
+++ b/aptos-core/consensus/src/block_storage/block_store.rs
@@ -653,7 +653,7 @@ impl BlockStore {
         if let Some(block_number) = block.block_number() {
             if !pipelined_block.has_randomness() {
                 if let Ok(Some(randomness)) = self.storage.consensus_db().get_randomness(block_number) {
-                    info!("Set randomness from DB for block {}, epoch {}, round {}", block_number, block.epoch(), block.round());
+                    debug!("Set randomness from DB for block {}, epoch {}, round {}", block_number, block.epoch(), block.round());
                     SET_RANDOMNESS_FROM_DB_COUNTER.with_label_values(&[]).inc();
                     pipelined_block.set_randomness(Randomness::new(
                         RandMetadata { epoch: block.epoch(), round: block.round() },
--- a/aptos-core/consensus/src/block_storage/sync_manager.rs
+++ b/aptos-core/consensus/src/block_storage/sync_manager.rs
@@ -476,7 +476,6 @@ impl BlockStore {
         info!("[FastForwardSync] Fetched {} blocks. Requested num_blocks {}. Initial block hash {:?}, target block hash {:?}",
             blocks.len(), num_blocks, highest_quorum_cert.certified_block().id(), highest_commit_cert.commit_info().id()
         );
-        info!("lightman1205 blocks: {:?}", blocks);
         for (i, (block, _)) in blocks.iter().enumerate() {
             assert_eq!(block.id(), quorum_certs[i].certified_block().id());
         }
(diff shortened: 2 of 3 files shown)


bin/gravity_node/src/reth_cli.rs
```diff
@@ -223,16 +225,25 @@ impl<EthApi: RethEthCall> RethCli<EthApi> {
...
+        
         loop {
+            let current_epoch = self.current_epoch.load(Ordering::SeqCst);
             // max executing block number
             let exec_blocks =
-                get_block_buffer_manager().get_ordered_blocks(start_ordered_block, None).await;
+                get_block_buffer_manager().get_ordered_blocks(start_ordered_block, None, current_epoch).await;
             if let Err(e) = exec_blocks {
                 let from = start_ordered_block;
                 if e.to_string().contains("Buffer is in epoch change") {
```
ByteYue: use one meaningful error type?

aptos-core/consensus/src/pipeline/buffer_manager.rs
```diff
@@ -411,11 +411,6 @@ impl BufferManager {
                     }))
                     .await
                     .expect("Failed to send persist request");
-                if commit_proof.ledger_info().ends_epoch() {
```
ByteYue: why delete these lines

crates/block-buffer-manager/src/block_buffer_manager.rs
```diff
@@ -337,8 +296,12 @@ impl BlockBufferManager {
...
     }
 
-    pub fn consume_epoch_change(&self) {
+    pub async fn consume_epoch_change(&self) -> (u64, u64) {
         self.buffer_state.store(BufferState::Ready as u8, Ordering::SeqCst);
+        let block_state_machine = self.block_state_machine.lock().await;
+        let current_epoch = block_state_machine.current_epoch;
+        let latest_epoch_change_block_number = *self.latest_epoch_change_block_number.lock().await;
+        (current_epoch, latest_epoch_change_block_number)
     }
```
keanji-x: return a struct with name

aptos-core/consensus/src/pipeline/buffer_manager.rs
Lchangliang: The buffer manager has been reset twice, once here and once in process_reset_request. process_reset_request is called when the epoch manager performs an epoch change. If this part is not removed, the block buffer manager will be released before the epoch change is done, and there will be hidden bugs

bin/gravity_node/src/reth_cli.rs
Lchangliang: Let's mention a separate PR later to modify this part

bin/gravity_node/Cargo.toml
```diff
@@ -32,7 +32,7 @@ serde = "^1.0.226"
 bincode = "1.3"
 time = "0.3.36"
 anyhow = "1.0.87"
-greth = { git = "https://github.com/Galxe/gravity-reth", rev = "b2bedd33a7130ad5cea096b781a0ca94b1518399" }
+greth = { git = "https://github.com/AshinGau/gravity-reth", rev = "c34cfe437" }
```
ByteYue: the source repo is not right.

PR #619, opened 2026-03-20
feat: non-blocking epoch change and fix deprecated try_next()

keanji-x:
## Summary

Implement non-blocking epoch change in the block buffer manager and fix deprecated `futures_channel` API usage.

### Non-blocking Epoch Change

Previously, when an epoch change block was detected, subsequent blocks in the same epoch ("suffix blocks") were dropped via `release_inflight_blocks()`, causing consensus to block waiting for execution results that would never arrive.

**Changes:**

1. **`block_buffer_manager.rs` — `set_compute_res()`**: When an epoch change is detected, immediately find all subsequent `Ordered` blocks in the same epoch and set `StateComputeResult::new_dummy()` as their compute result. This allows consensus to consume these results instantly without blocking.

2. **`block_buffer_manager.rs` — `get_committed_blocks()`**: Skip suffix blocks past the `latest_epoch_change_block_number` boundary. These blocks have dummy execution results and were never executed by reth, so they must not enter the reth commit path.

### Fix deprecated `try_next()` → `try_recv()`

Replace all deprecated `futures_channel::mpsc::UnboundedReceiver::try_next()` calls with `try_recv()` across the consensus crate.

### Dependency Updates

- Update `gaptos` to branch `feat/add-epoch-block-info` (Galxe/gravity-aptos#57)
- Update `greth` to branch `feat/add-epoch-block-info` (Galxe/gravity-reth#305)

## Related PRs

- Galxe/gravity-aptos#57
- Galxe/gravity-reth#305

### DIFF
--- a/aptos-core/consensus/src/block_storage/block_store.rs
+++ b/aptos-core/consensus/src/block_storage/block_store.rs
@@ -666,6 +666,25 @@ impl BlockStore {
                 commit_decision,
             );
         } else {
+            // `BATCH_COMMIT_SIZE` env var: when set and the accumulated path is still
+            // small, defer sending so the execution layer can process blocks in larger
+            // batches. Because we return early without touching `ordered_root`, the next
+            // `send_for_execution` call will see the same root and `path_from_ordered_root`
+            // will grow until it exceeds the batch size, at which point we flush.
+            // Used by `gravity_e2e/cluster_test_cases/single_node/test_batch_exec.py`.
+            let batch_commit_size = std::env::var("BATCH_COMMIT_SIZE")
+                .ok()
+                .and_then(|v| v.parse::<usize>().ok())
+                .unwrap_or(0);
+            if batch_commit_size > 0 && blocks_to_commit.len() <= batch_commit_size {
+                info!(
+                    "blocks_to_commit len {} <= BATCH_COMMIT_SIZE {}, defer for batch accumulation",
+                    blocks_to_commit.len(),
+                    batch_commit_size
+                );
+                return Ok(());
+            }
+
             info!("send the blocks to execution {:?}", blocks_to_commit);
             self.execution_client
                 .finalize_order(
--- a/aptos-core/consensus/src/block_storage/block_store.rs
+++ b/aptos-core/consensus/src/block_storage/block_store.rs
@@ -630,7 +630,7 @@ impl BlockStore {
                     enable_randomness: self.enable_randomness,
                 };
                 get_block_buffer_manager()
-                    .set_ordered_blocks(BlockId(*p_block.parent_id()), block)
+                    .set_ordered_blocks(BlockId(*p_block.parent_id()), block, p_block.round())
                     .await
                     .context("Failed to set ordered blocks during recovery")?;
                 let compute_res = get_block_buffer_manager()
--- a/aptos-core/consensus/src/block_storage/block_store.rs
+++ b/aptos-core/consensus/src/block_storage/block_store.rs
@@ -229,35 +229,70 @@ impl BlockStore {
         LedgerInfoWithSignatures::new(LedgerInfo::dummy(), AggregateSignature::empty())
     }
 
+    /// Replays uncommitted quorum certificates after a node restart.
+    ///
+    /// Retrieves all QCs with commit info from the block tree, sorts them by round,
+    /// and re-sends each one for execution in order. This reproduces the same commit
+    /// batches that were in-flight before the restart.
+    ///
+    /// If the latest ledger info carries `epoch_block_info` (indicating a non-blocking
+    /// epoch change was in progress), recovery stops at the epoch change block's round.
+    /// Suffix blocks beyond that point would never receive execution results from reth,
+    /// so attempting to recover them would cause the pipeline to hang.
     async fn recover_blocks(&self) {
         RECOVERY_GAUGE.set_with(&[], 1);
-        // reproduce the same batches (important for the commit phase)
+
         let mut certs = self.inner.read().get_all_quorum_certs_with_commit_info();
         let last_ledger_info =
             self.storage.consensus_db().ledger_db.metadata_db().get_latest_ledger_info();
-        info!("recover blocks, last_ledger_info: {:?}", last_ledger_info);
+        info!("recover_blocks: last_ledger_info={:?}", last_ledger_info);
+
+        // Determine the upper-bound round for recovery when a non-blocking epoch change
+        // was in progress. Blocks after this round are suffix blocks that reth will discard.
+        let epoch_change_limit_round =
+            last_ledger_info.as_ref().and_then(|li| {
+                let info = li.ledger_info().commit_info().epoch_block_info()?;
+                let block = self.get_block(info.block_id)?;
+                info!(
+                "recover_blocks: epoch change detected at block_id={}, block_number={}, round={}",
+                info.block_id, info.block_number, block.round(),
+            );
+                Some(block.round())
+            });
+
         certs.sort_unstable_by_key(|qc| qc.commit_info().round());
+
         for qc in certs {
-            if qc.commit_info().round() > self.commit_root().round() {
-                if let Some(last_ledger_info) = &last_ledger_info {
-                    if last_ledger_info.ledger_info().epoch() == qc.commit_info().epoch() &&
-                        qc.commit_info().round() <= last_ledger_info.commit_info().round()
-                    {
-                        info!(
-                            "sending qc {} to execution, current commit round {}",
-                            qc.commit_info().round(),
-                            self.commit_root().round()
-                        );
-                        if let Err(e) =
-                            self.send_for_execution(qc.into_wrapped_ledger_info(), true).await
-                        {
-                            error!("Error in try-committing blocks. {}", e.to_string());
-                            break;
-                        }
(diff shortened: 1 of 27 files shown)


ByteYue (dismissed):

Lchangliang (dismissed):

keanji-x (commented):
Overall the non-blocking epoch change approach is solid. Left a few inline comments on specific concerns.

crates/block-buffer-manager/src/block_buffer_manager.rs
```diff
@@ -731,6 +829,70 @@ impl BlockBufferManager {
...
+            });
+            info!(
+                "[EpochChange] set_compute_res: suffix block {:?} num {:?} epoch {:?} \
+                 (dummy result, epoch change at block {})",
+                suffix_block_id, suffix_key.block_number, suffix_key.epoch, epoch_change_block_num,
+            );
+        }
+
+        if !suffix_keys.is_empty() {
+            info!(
```
keanji-x, line 906: Suffix blocks that are already `Ordered` when the epoch change is detected get `StateComputeResult::new_dummy()` here, but suffix blocks that arrive later via `get_executed_res` get `epoch_change_block_info.compute_result.clone()` — the actual epoch change block's result, which has `has_reconfiguration() == true`.

Same logical category of blocks, different compute results. Downstream code checking `has_reconfiguration()` could behave differently depending on timing. Should both paths use `new_dummy()` consistently?

crates/block-buffer-manager/src/block_buffer_manager.rs
```diff
@@ -352,15 +380,42 @@ impl BlockBufferManager {
...
+        // release_inflight_blocks. Only now signal that the epoch change has been consumed
         self.buffer_state.store(BufferState::Ready as u8, Ordering::SeqCst);
-        epoch
+
+        // If epoch_change_block_number is 0, it means this is a redundant call
+        // (release_inflight_blocks can set buffer_state=EpochChange again after
+        // a previous consume). Use latest_commit_block_number as fallback to
+        // prevent the caller from resetting start_ordered_block to 1.
+        let effective_block_number = if epoch_change_block_number == 0 {
+            block_state_machine.latest_commit_block_number
```
keanji-x, line 424: After `consume_epoch_change` resets `latest_epoch_change_block_number` to 0:
- `is_suffix_block()` still returns `true` (checks `epoch_change_block_info` directly)
- `get_epoch_change_block_info()` returns `None` (requires `epoch_change_bn > 0`)

This asymmetry means the reth side and consensus side disagree on what's a suffix block during the window between `consume_epoch_change` and `release_inflight_blocks`. If consensus calls `resolve_epoch_block_info` in this window, it won't find the `EpochBlockInfo` it needs.

Consider making both methods use the same predicate, or document why the divergence is safe.

aptos-core/consensus/src/block_storage/block_store.rs
```diff
@@ -630,83 +665,30 @@ impl BlockStore {
...
-                    // Last block: use the original finality_proof
-                    finality_proof.clone()
-                } else {
-                    // Look up the QC certifying the next block in the chain.
-                    // That QC's commit_info should point to this block (2-chain rule).
-                    let next_qc = (i + 1 < blocks_to_commit.len())
-                        .then(|| self.get_quorum_cert_for_block(blocks_to_commit[i + 1].id()))
-                        .flatten();
-                    match next_qc {
-                        Some(qc) if qc.commit_info().id() == block.id() => {
```
keanji-x, line 665: The block-by-block commit logic with `BATCH_COMMIT_SIZE` is removed here but not mentioned in the PR description. Is this intentional? Any production deployments relying on that env var?

aptos-core/consensus/src/block_storage/sync_manager.rs
```diff
@@ -544,44 +544,44 @@ impl BlockStore {
             ledger_info.commit_info().round()
         );
 
-        // if the block exists between commit root and ordered root
+        // Step 1: If the block already exists between commit root and ordered root,
```
keanji-x, line 547: The `has_missing_randomness` guard was removed from this condition. Previously missing randomness would skip the "send commit proof" path and go straight to full sync. Now it always sends the commit proof first, then potentially also syncs. Intentional change or side effect of the refactor?

aptos-core/consensus/src/pipeline/buffer_manager.rs
```diff
@@ -256,9 +261,52 @@ impl BufferManager {
...
+                warn!(
+                    round = round,
+                    block_id = block_id,
+                    "Too many pending commit proofs, ignored."
+                );
+                false
+            }
+        } else {
+            debug!(
+                round = round,
```
keanji-x, line 307: When the cache hits `MAX_PENDING_COMMIT_PROOFS`, new proofs are rejected. Since older proofs are more likely stale, consider evicting the oldest entry (`pop_first()` from the BTreeMap) instead of refusing new ones.

keanji-x (commented):
Overall the non-blocking epoch change approach is solid — returning dummy results for suffix blocks instead of dropping them is the right call. Left a few comments on specific concerns.

crates/block-buffer-manager/src/block_buffer_manager.rs
Lchangliang: Fixed — both paths now use `StateComputeResult::new_dummy()` consistently. The old-epoch bypass branch in `get_executed_res` got the same treatment, and the now-unused `compute_result` field on `EpochChangeCache` was removed so future consumers can't accidentally depend on it.

crates/block-buffer-manager/src/block_buffer_manager.rs
Lchangliang: Good catch. Fixed — `get_epoch_change_block_info` now uses `epoch_change_block_info` as its source of truth (same predicate as `is_suffix_block`), so the reth and consensus sides agree across the `consume_epoch_change` → `release_inflight_blocks` window. Also rewrote the stale comment on `consume_epoch_change` — the old note claimed the reset was for `get_committed_blocks`, but that function never checked this field; the actual reason is `release_inflight_blocks` retain semantics.

aptos-core/consensus/src/block_storage/block_store.rs
Lchangliang: Intentional, sorry for the missing context. `BATCH_COMMIT_SIZE` is only consumed by `gravity_e2e/cluster_test_cases/single_node/test_batch_exec.py` — no production deployment uses it. I've re-added the accumulation logic in a simpler form: `finalize_order` can already commit an N-block batch with just the last block's proof, so we just early-return from `send_for_execution` while the accumulated path is under the threshold and let `path_from_ordered_root` grow on the next call. Default behavior (env var unset) is unchanged.

aptos-core/consensus/src/block_storage/sync_manager.rs
Lchangliang: Intentional — we need to guarantee the commit vote is sent regardless of randomness state. Will update the PR description to call this out.

aptos-core/consensus/src/pipeline/buffer_manager.rs
Lchangliang: Agreed, done. The cache now evicts the oldest pending proof via `pop_first()` when full — but only if the incoming proof is strictly newer than the oldest, otherwise we'd drop a newer proof to accept an older one.

Lchangliang (dismissed):

aptos-core/consensus/src/block_storage/sync_manager.rs
keanji-x, line 547: This is the one remaining item from the previous review. The old code skipped the "send commit proof" branch entirely when `has_missing_randomness` was true, forcing a full sync. Now Step 1 always fires regardless of `has_missing_randomness`, and execution falls through to Step 2.

Is this intentional? If so, could you add a brief comment explaining why it's safe to send the commit proof even when randomness data is missing?

keanji-x (commented):
Re-reviewed after fix commit `e91d0f3d`. 4 out of 5 previous issues are addressed — nice work.

**Resolved:**
1. **Inconsistent dummy results** — `compute_result` removed from `EpochChangeCache`, all suffix block paths now use `StateComputeResult::new_dummy()` consistently. LGTM.
2. **`is_suffix_block` vs `get_epoch_change_block_info` asymmetry** — `get_epoch_change_block_info` now uses `epoch_change_block_info` directly instead of `latest_epoch_change_block_number`, matching `is_suffix_block`. LGTM.
3. **`BATCH_COMMIT_SIZE` removal** — Re-added with clear comment referencing `test_batch_exec.py`. LGTM.
4. **`pending_commit_proofs` eviction** — Now evicts oldest entry via `pop_first()` with a guard against inserting older-than-oldest proofs. LGTM.

**Still open:**
5. **`has_missing_randomness` guard removal in `sync_to_highest_commit_cert`** — not addressed in the fix commit. Left an inline comment.

Lchangliang (dismissed):

Lchangliang (dismissed):

Lchangliang (approved):

Lchangliang (dismissed):

Lchangliang (approved):

PR #639, opened 2026-04-06
feat(consensus): add failed_proposer_indices and validator performance based leader election

Lchangliang:
- Add failed_proposer_indices to ExternalBlockMeta in block_store, pipeline_builder, and state_computer
- Switch leader reputation weights from history-based to ValidatorPerformances from EVM config storage
- Simplify epoch proposer extraction (skip cross-epoch history lookup)
- Fix chain health ratio to handle empty history (prevent halt at Epoch 3)
- Rename auto_evict_threshold to auto_evict_threshold_pct in genesis configs
- Update gaptos/greth deps to dev-0402-performance branch, rust toolchain to 1.93.0
- Add ValidatorPerformances to config storage allowed configs


keanji-x (dismissed):

keanji-x (commented):
## PR Review

### Critical Issues

#### 1. Backend completely bypassed — chain health monitoring is dead code
`leader_reputation.rs:718-721` hardcodes `sliding_window = vec![]` and `root_hash = HashValue::zero()`, which means:
- `compute_chain_health_and_add_metrics` always receives empty history → always returns `1.0`
- All Prometheus metrics (`CHAIN_HEALTH_*`) will report zeros
- `use_root_hash` is meaningless since root_hash is always zero

If this is intentional, the `backend` field, `window_for_chain_health`, and `compute_chain_health_and_add_metrics` should be cleaned up rather than left as dead code.

#### 2. Per-round `ValidatorPerformances` fetch in hot path
`get_weights` is called every round during leader election. Each call does:
1. `GLOBAL_CONFIG_STORAGE.get()`
2. `fetch_config_bytes(ValidatorPerformances, Latest)`
3. BCS deserialization

This could become a performance bottleneck at scale. Consider caching with a TTL or refreshing per-epoch.

#### 3. New `block_event` schema file is unused
`consensusdb/schema/block_event/mod.rs` (51 lines) defines `BlockEventSchema` but nothing imports or references it. Should be removed or wired in.

#### 4. Inconsistent `failed_proposer_indices` extraction across 3 call sites
- `block_store.rs`: uses `self.validator_indices.get(author)` (HashMap lookup)
- `pipeline_builder.rs` & `state_computer.rs`: use `Block::failed_authors_to_indices()` (linear scan)

Different implementations could produce divergent results. Should use one consistent approach.

### Minor Issues

- **MOCK HACK comment in production code** (`leader_reputation.rs:686-688`): The comment references mocking but this is now the permanent behavior. Should be updated.
- **`_history` and `_needed_rounds` dead parameters**: If history-based election is permanently abandoned, consider cleaning up the trait interface.
- **Rust toolchain 1.88.0 → 1.93.0**: Large jump — ensure CI passes with new clippy lints.
- **`auto_evict_threshold` → `auto_evict_threshold_pct`**: Breaking change for existing genesis configs. Ensure migration is documented.
- **3 out of 4 commits are named "tmp"**: Please squash before merge.

Lchangliang:
Thanks for the review! Here's the status on each point:

### Critical Issues

**1. Backend bypassed / chain health dead code**
Acknowledged. `get_block_metadata` is Aptos-native and not applicable to gsdk/greth — we use `ValidatorPerformances` from EVM config storage instead. Added `NOTE(gravity)` comments to `compute_chain_health_and_add_metrics` and the `block_event` schema explaining this. Not removing the code to minimize diff against upstream Aptos.

**2. Per-round `ValidatorPerformances` fetch in hot path**
Fixed. Added a cache with `PERFORMANCE_CACHE_REFRESH_ROUNDS = 100` — weights are now fetched once and reused for 100 rounds before refreshing.

**3. `block_event` schema unused**
Same as #1 — this is Aptos-native code kept for compatibility. Added a `NOTE(gravity)` comment to the file.

**4. Inconsistent `failed_proposer_indices` extraction**
Fixed. Unified all three call sites (`block_store.rs`, `pipeline_builder.rs`, `state_computer.rs`) to use the same `filter_map` style that silently skips unknown authors instead of panicking.

### Minor Issues

- **MOCK HACK comment**: Fixed — updated to accurately describe the Gravity behavior.
- **`_history` / `_needed_rounds` dead parameters**: Not changing — these are from the upstream Aptos trait interface. Modifying them would increase merge friction with upstream for no practical benefit.
- **Rust toolchain 1.88→1.93**: CI clippy issue fixed (`derivable_impls` in sentinel config).
- **`auto_evict_threshold` rename**: Noted, will document in release notes.
- **tmp commits**: Will squash before merge.

Also lowered the per-validator weight assignment logs from `info!` to `debug!` to reduce log noise, and cleaned up the `gaptos::aptos_logger::` prefix (already imported via `prelude::*`).

keanji-x (dismissed):

ByteYue (dismissed):

keanji-x (dismissed):

ByteYue (dismissed):

ByteYue (approved):

PR #652, opened 2026-04-14
fix(consensus): address 4 tier-1 stability issues from audit

keanji-x:
## Summary

Fixes 4 production stability bugs identified in the [gravity-audit](https://github.com/Galxe/gravity-audit/issues):

- **Galxe/gravity-audit#62** — Infinite loop in `push_bucket_to_batches` when oversized tx heads the queue → skip with warning
- **Galxe/gravity-audit#54** — `tokio::select!` silently drops persist-phase errors → log the error
- **Galxe/gravity-audit#55** — `BufferManager::reset()` spin loop has no timeout → 30s deadline to prevent deadlock
- **Galxe/gravity-audit#65** — `.expect()` on channel sends cascades panic across quorum store → graceful error + break

Closes https://github.com/Galxe/gravity-audit/issues/62
Closes https://github.com/Galxe/gravity-audit/issues/54
Closes https://github.com/Galxe/gravity-audit/issues/55
Closes https://github.com/Galxe/gravity-audit/issues/65

These are the "Tier 1" issues from audit triage — bugs that can be triggered without any malicious behavior (user submitting large tx, disk hiccup, task leak, component crash).

## Changes

| File | Fix |
|------|-----|
| `quorum_store/batch_generator.rs` | Handle `num_batch_txns == 0` case: drain oversized tx, warn, continue |
| `pipeline/buffer_manager.rs` | Add `Some(Err(e))` arm to persisting_phase select |
| `pipeline/buffer_manager.rs` | Add `Instant::now() + 30s` deadline to reset spin loop |
| `quorum_store/network_listener.rs` | `.expect()` → `.is_err()` + `error!()` + `break` |

## Test plan

- [ ] Verify existing consensus tests pass
- [ ] Manual review of each fix for correctness
- [ ] The pre-existing `disable_lifo_slot` build error in upstream `gravity-aptos` dependency is unrelated

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #672, opened 2026-04-21
feat(gravity_cli): add --kms flag for Cloud KMS-backed EVM signing

keanji-x:
## Summary

Adds an opt-in `--kms <resource>` flag to `validator join`, `validator leave`, and `stake create` so they can sign EVM transactions with a key held in Google Cloud KMS instead of a plaintext private key. The existing interactive stdin prompt is preserved as the default — no behavioral change unless `--kms` is passed.

This is the gravity_cli side of the infrastructure work landing under `mono-grav/scripts/kms/` (separate repo): there, per-node HSM keys + service accounts + IAM bindings are provisioned. Here, the CLI learns to actually consume them.

## What changes

New `signer` module under `bin/gravity_cli/src/signer/`:

- `SignerArgs` — clap `Args` struct with a single `--kms` flag. Flattened into the three command structs.
- `SignerArgs::resolve() -> ResolvedSigner` — returns an `EthereumWallet` ready for `ProviderBuilder::wallet()` plus the derived address. Two branches:
  - `--kms` → `GcpKmsSigner` (network call to fetch pubkey, then derive address)
  - no flag → existing `rpassword` stdin prompt
- `GcpKmsSigner` — implements `alloy_signer::Signer` and `alloy_network::TxSigner<Signature>`. Handles DER → (r,s), EIP-2 low-S normalization, and recovery-id brute force against the cached public key.

Three call sites updated identically (delete the `prompt_password_stdout` block, replace with `let resolved = self.signer.resolve().await?`):

- `bin/gravity_cli/src/validator/join.rs`
- `bin/gravity_cli/src/validator/leave.rs`
- `bin/gravity_cli/src/stake/create.rs`

`Cargo.toml`: adds `google-cloud-kms = "0.6"`, `async-trait = "0.1"`, `alloy-network` and `alloy-consensus` (needed for the signer trait impls), and the `pem` feature on `k256` (needed to parse the KMS-returned public key).

## Why no `--private-key-env`

An earlier revision of this PR exposed a `--private-key-env <VAR>` option that read the hex key from an environment variable, intended for CI/automation. It was removed: a plaintext key in an env var is visible in `/proc/<pid>/environ`, can leak into shell history or CI logs, and provides no real security advantage over the stdin prompt — it only adds attack surface. If non-interactive signing is needed, use `--kms`.

## Usage

Existing flow (unchanged):

```
gravity_cli validator join --rpc-url http://… --stake-pool 0x… …
# prompts: Enter private key
```

KMS flow (new):

```
gravity_cli validator join --rpc-url http://… --stake-pool 0x… \
  --kms projects/gravity-infrastructure/locations/us-west1/keyRings/gravity-mainnet/cryptoKeys/node1
```

(The trailing `/cryptoKeyVersions/N` is optional; defaults to version `1`.)

## Test plan

- [x] `cargo fmt -p gravity_cli` (nightly, matches workspace `rustfmt.toml`)
- [x] `cargo clippy -p gravity_cli --all-targets -- -D warnings`
- [x] `cargo test -p gravity_cli signer::` — address-derivation vector test (Anvil account #0) + KMS resource-path normalization tests pass
- [ ] `gravity_cli validator join …` (no flag) — confirm the stdin prompt still works as before
- [ ] `gravity_cli validator join … --kms projects/…/cryptoKeys/node1` against a real KMS key on testnet — confirm the join tx is accepted and recovers to the address derived from the KMS public key

## Notes for review

- The KMS signer body has only been smoke-tested via unit tests on pure functions (address derivation, path normalization). DER parsing, S normalization, and recovery-id resolution have not been exercised against a live KMS — please verify on testnet before relying on this flag in mainnet.
- The `wallet_address` lookup is now async (`resolve().await?`) since the KMS path needs to fetch the public key. The non-KMS path completes synchronously inside the future.
- Re: the review comment on `DigestVariant::Sha256` — the code has an inline comment explaining the convention. In short: the `digest` oneof on GCP KMS is a **type tag** matching the key algorithm; KMS signs the 32 bytes as-is and does NOT re-hash. Re-hashing would only happen via the sibling `data` field. This mirrors alloy-signer-aws's pattern for Ethereum signing. Still worth verifying on testnet per the test-plan checklist.


ByteYue (dismissed):

Lchangliang (commented):
## Review Notes

### 1. Cargo.toml version mismatch (Bug)

PR description says `google-cloud-kms = "0.7"`, but the actual Cargo.toml adds `"0.6"`. These should be reconciled — confirm which version is intended and update whichever is wrong.

### 2. `sign_hash` passes a pre-hashed digest into `DigestVariant::Sha256` (Potential Bug — Critical)

```rust
// kms.rs:113
digest: Some(Digest {
    digest: Some(DigestVariant::Sha256(hash.as_slice().to_vec())),
}),
```

`Signer::sign_hash` receives `hash` that is already the result of `keccak256` (32 bytes). Wrapping it in `DigestVariant::Sha256` tells KMS "this is a SHA-256 digest" — but it is actually keccak256.

**Key concern**: if the KMS key is configured with digest algorithm `SHA256` (i.e. `EC_SIGN_SECP256K1_SHA256`), KMS will SHA-256 the input *again* before signing, producing a signature over `SHA256(keccak256(msg))` instead of `keccak256(msg)`. That signature will fail EVM `ecrecover`.

The correct approach is either:
- Use the `data` field on `AsymmetricSignRequest` (raw signing, available with `EC_SIGN_SECP256K1_SHA256` keys since KMS v1), or
- Ensure the KMS key algorithm is one that accepts a raw pre-hashed digest without re-hashing.

**This is the most important item to verify on testnet before merge.** If the key type doesn't match the signing path, every transaction will silently produce an invalid signature.

### 3. `chain_id` is never injected into `GcpKmsSigner` (Functional gap)

```rust
// kms.rs:83
Ok(Self { client, key_resource, address, verifying_key, chain_id: None })
```

`chain_id` is always `None`, and `mod.rs:76` never calls `set_chain_id` after construction. The `PrivateKeySigner` path works because `EthereumWallet` / `ProviderBuilder` handles chain_id injection automatically for well-known signer types — but for a custom `Signer` impl this may not happen. Worth confirming that EIP-1559 transaction signing works correctly without an explicit chain_id on the signer, or add a `set_chain_id` call after construction using the provider's chain_id.

### 4. `--private-key-env`: security caveat worth documenting

The private key via env var is visible in `/proc/<pid>/environ` (Linux), and inline usage like `GRAV_KEY=xxx gravity_cli ...` may leak into shell history. Since the whole point of this PR is to move toward KMS and away from plaintext keys, consider adding a brief security note in `--help` text (e.g. "Prefer --kms for production; this flag is intended for CI/automation where the env var is injected securely").

### 5. KMS auth error message could be more actionable

```rust
.map_err(|e| anyhow::anyhow!("KMS auth setup failed: {e}"))?;
```

When `GOOGLE_APPLICATION_CREDENTIALS` is unset or the metadata server is unreachable, this produces a generic error. A hint like "check Application Default Credentials configuration" would save debugging time.

### 6. Duplicated doc comment on `signer` field across three call sites

The `/// Signing-key source: defaults to interactive stdin prompt...` block is copy-pasted identically in `join.rs`, `leave.rs`, and `create.rs`. Minor, but easy to drift — since the docs live on `SignerArgs` itself, the per-field copies could be shortened or removed.

---

### Summary

The architecture is clean and the call-site refactor is mechanical and correct. **The critical item is #2** — the SHA256/keccak256 double-hashing risk. If the KMS key's algorithm causes re-hashing, every signature will be invalid on-chain. This must be verified against a real KMS key on testnet before merge. Everything else is minor.

keanji-x:
Thanks for the careful read — responses point-by-point.

### 1. Cargo.toml version mismatch — fixed

PR description now says `0.6` to match Cargo.toml. `0.7` hasn't been published to crates.io yet (latest is 0.6.0); the description was written optimistically.

### 2. SHA256 / keccak256 double-hashing — verified, not a bug (but the comment needed work)

The reviewer's concern is real if the code had used the `data` field. But we're using the `digest` oneof, and on GCP KMS `digest.sha256/sha384/sha512` is a **type tag** matching the key algorithm — KMS signs the 32 bytes as-is and does **not** re-hash. Re-hashing only happens via the sibling `data` field. This is the same pattern [`alloy-signer-aws`](https://github.com/alloy-rs/alloy/tree/main/crates/signer-aws) uses: pass the keccak256 output as the \"SHA-256\" digest, the KMS ECDSA signer doesn't care that the digest wasn't produced by SHA-256.

Code unchanged but I added a long inline comment at `signer/kms.rs:109-117` explaining this, so the next reader doesn't have to re-derive it. Commit [`a2630a386`](https://github.com/Galxe/gravity-sdk/pull/672/commits/a2630a386).

Still flagged in the PR test plan as requiring a real testnet `validator join --kms …` to verify end-to-end — that's the only test that catches an algorithm / field-name mix-up with certainty.

### 3. `chain_id` always `None` — intentional, now documented

`sign_hash` signs a pre-computed 32-byte digest (`tx.signature_hash()`), which already folds the chain_id in via EIP-155 — the signer's internal `chain_id` is never consumed in the sign path. I've kept the field and `set_chain_id` to honour the `Signer` trait contract (some alloy versions may call it defensively), but it's informational only. Added a doc comment on the field at `signer/kms.rs:52-57` making this explicit. Commit [`a2630a386`](https://github.com/Galxe/gravity-sdk/pull/672/commits/a2630a386).

### 4. `--private-key-env` — removed entirely

Your suggested caveat was \"prefer `--kms` for production; `--private-key-env` is intended for CI/automation\". After a second look: a plaintext key in an env var provides **no real security advantage** over the stdin prompt — it is visible in `/proc/<pid>/environ`, can leak into shell history or CI logs — and the mere existence of a second plaintext path widens the attack surface. If non-interactive signing is needed, the right answer is `--kms`, not a differently-packaged plaintext input.

So `--private-key-env` is gone. `SignerArgs` now has a single flag (`--kms`) and a single fallback (stdin prompt). Commit [`60a6789bd`](https://github.com/Galxe/gravity-sdk/pull/672/commits/60a6789bd).

### 5. Auth error message — now actionable

The generic \"KMS auth setup failed\" is replaced with:

> KMS auth setup failed: <err> — check Application Default Credentials: on GCE the VM's service account must be bound with cloud-platform scope; locally set GOOGLE_APPLICATION_CREDENTIALS or run `gcloud auth application-default login`

`signer/kms.rs:78-84`. Commit [`a2630a386`](https://github.com/Galxe/gravity-sdk/pull/672/commits/a2630a386).

### 6. Duplicated doc comment on the `signer` field — removed

Dropped the copy-pasted `/// Signing-key source: ...` block from `join.rs`, `leave.rs`, and `stake/create.rs`. `SignerArgs` has a single canonical docstring; the three call sites now just flatten it. Commit [`a2630a386`](https://github.com/Galxe/gravity-sdk/pull/672/commits/a2630a386).

---

### Summary

- **#1, #5, #6** straightforwardly fixed.
- **#2** is a documentation-only change — the code was correct (same pattern as `alloy-signer-aws`) but the reasoning was not obvious from reading it. Now spelled out in a comment.
- **#3** is a deliberate design choice — the signer's `chain_id` is trait etiquette, not load-bearing. Documented.
- **#4** was resolved by going further than the original suggestion and **removing** `--private-key-env`. Thanks for highlighting the attack-surface concern — it's what prompted the re-think.

Still needs a testnet `--kms` run before merge to close out the digest / DER / recovery-id chain with real bytes.

Lchangliang (approved):

keanji-x:
Superseded by #683 — rolled up together with #678 and #652 into a single combined PR to cut CI runs. Closing here; all 4 commits from this branch are cherry-picked onto the new branch with no conflicts.

PR #674, opened 2026-04-21
fix(consensus): drop RSA JWK updates with error log instead of forwarding

ByteYue:
## Summary
- RSA JWKs are never expected to reach `state_computer.rs` in this system (JWKs originate from the external `aptos-jwk-consensus` runtime, but we do not plan to ever surface RSA to the execution layer).
- Previously both `process_jwk_update` and `process_jwk_update_util` serialized `JWK::RSA` into `ExtraDataType::JWK` and forwarded it to the execution layer — no filtering, no visibility if one ever slipped through.
- Change: the RSA arm in both functions now logs an `error!` (with block id, issuer, version, kid) and filters the entry out. `process_jwk_update_util`'s `.map()` becomes `.filter_map()` so the drop is actually applied. The now-unused `JWK_TYPE_RSA` constant is removed.

Scope is deliberately narrow — this is a boundary-level firewall inside state_computer, not a change to upstream `aptos-jwk-consensus`. If the upstream ever produces an RSA JWK, it is now contained here and visible in logs.

## Test plan
- [x] `cargo check -p aptos-consensus` (local) passes.
- [ ] Run an E2E cluster and confirm no RSA JWK is observed in normal operation (no error logs fired).
- [ ] Optional: inject a synthetic RSA JWK via test and confirm the error log fires and the JWK is dropped from the serialized `ExtraDataType::JWK`.

### DIFF
--- a/aptos-core/consensus/src/state_computer.rs
+++ b/aptos-core/consensus/src/state_computer.rs
@@ -61,7 +61,6 @@ use gaptos::{
 use std::{boxed::Box, iter::once, sync::Arc, time::Duration};
 
 /// JWK type name constants for consistent usage across SDK ↔ reth boundary
-const JWK_TYPE_RSA: &str = "0x1::jwks::RSA_JWK";
 const JWK_TYPE_UNSUPPORTED: &str = "0x1::jwks::Unsupported_JWK";
 use tokio::sync::Mutex as AsyncMutex;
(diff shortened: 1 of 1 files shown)


AshinGau:
LGTM

AshinGau (approved):

PR #680, opened 2026-04-22
feat(docker): binary-only image + compose flows for validator deployment

keanji-x:
## Summary

- Replace the legacy `docker/gravity_node/` — its Dockerfile still referenced the `deploy_utils/` directory that was removed in #538, so it hadn't built for months. The old `config/` also carried hard-coded, fixed-per-node identity and waypoint files.
- New multi-stage Dockerfile produces a binary-only image (`rust:1.93-slim` builder → `debian:12-slim` runtime, non-root uid `10001`, `tini` as PID 1). Genesis, waypoint, identity, and yaml configs are mounted from the host at runtime so the same image ships to every node.
- Two compose flows:
  - `docker-compose.yaml` — single-validator production layout. Config dir read-only mount, chain data on named volumes so upgrades are `docker compose up -d` against a new tag without touching state.
  - `docker-compose.cluster.yaml` — 4 validators + 1 VFN on one host, for end-to-end image verification against `cluster/` artifacts. `network_mode: host` on every service (genesis-encoded peer addrs require it).
- `render-cluster-config.sh` regenerates the 5-node config set from `cluster/output/` + `cluster/templates/`, rewriting paths to the container-internal locations.
- `README.md` documents build, devnet verification, single-node deployment, registry-less distribution via `docker save` + `scp`, upgrade flow, and security notes.
- Repo-root `.dockerignore` keeps `target/` and other large paths out of the build context.

## Verified end-to-end

Locally ran the 4+1 topology to confirm the image reaches consensus:

- Image built with `CARGO_PROFILE=release` against current `main`.
- Configs rendered via `render-cluster-config.sh` from `cluster/output/`.
- `docker compose -f docker-compose.cluster.yaml up -d` — all five containers healthy.
- Block height advanced from `0x4ba` to `0x53d` over ~20 s across every node. VFN tracked one block behind the validators. Consensus logs showed commit votes from three distinct validator keys per round.

## Test plan

- [ ] `DOCKER_BUILDKIT=1 docker build --network=host -f docker/gravity_node/Dockerfile -t gravity_node:ci --build-arg CARGO_PROFILE=release .` succeeds.
- [ ] `cd cluster && make init && make genesis` produces artifacts under `cluster/output/`.
- [ ] `cd docker/gravity_node && ./render-cluster-config.sh` populates `config/{node1..node4,vfn1}/`.
- [ ] `IMAGE_TAG=ci docker compose -f docker-compose.cluster.yaml up -d` brings up all 5 services.
- [ ] `eth_blockNumber` on ports `8545..8548` and `8550` returns an increasing hex block number across successive calls.
- [ ] `docker compose -f docker-compose.cluster.yaml down -v` tears down cleanly and frees the port range.

## Follow-ups (out of scope for this PR)

- Mainnet-specific config templates (strip `dev`, replace loopback hosts with real addresses, set real chain id).
- Registry choice (GHCR / ECR / Harbor) and CI publish workflow.
- Integrate `gravity_cli --kms` once the signer branch lands — drops the on-disk `identity.yaml` entirely.
- Strip debug symbols from the release binary (image is ~6.7 GB as built; stripping or switching to `performance` profile cuts this sharply).


Richard1048576 (approved):
LGTM

PR #697, opened 2026-04-29
fix(consensus): use block_number limit in recover_blocks to prevent skipping epoch change

Lchangliang:
## Summary

Fix two recovery panics that prevent nodes from restarting after crashes:

**1. recover_blocks skips epoch change block (commits 541e93eb, 5d9fc047, fd685e7d)**

The round-based `epoch_change_limit_round` check in `recover_blocks` would break before processing the only QC that can commit the epoch change block, because that QC's `commit_round` exceeds the epoch change round. Replace with a `block_number`-based limit passed into `send_for_execution`, so the epoch change block is committed while suffix blocks are skipped.

**2. "No LI found for root" panic on implicit-only commit (commit a28eeaa4)**

When reth lags behind aptos_db (e.g. crash between aptos commit and reth persist), the recovery root block may have been committed implicitly as an ancestor via the 3-chain rule, with no explicit commit LI in consensus DB (e.g. due to consecutive timeouts after the root round). This caused `find_root_by_block_number` to panic. Fix: construct a placeholder `WrappedLedgerInfo` with the root block's `BlockInfo` to pass `BlockTree::new`'s assert, then let `recover_blocks()` replay the real commit proofs via `commit_callback` and overwrite it.

## Test plan

- [ ] Verify node recovers from crash where reth block_number < aptos_db block_number
- [ ] Verify node recovers across epoch change boundaries
- [ ] Verify normal restart (with explicit commit LI) still works unchanged

### DIFF
--- a/aptos-core/consensus/src/block_storage/block_store.rs
+++ b/aptos-core/consensus/src/block_storage/block_store.rs
@@ -249,15 +249,14 @@ impl BlockStore {
 
         // When a non-blocking epoch change was in progress, determine the epoch change
         // block_number so that recovery does not commit suffix blocks past it.
-        let epoch_change_block_number =
-            last_ledger_info.as_ref().and_then(|li| {
-                let info = li.ledger_info().commit_info().epoch_block_info()?;
-                info!(
-                    "recover_blocks: epoch change detected at block_id={}, block_number={}",
-                    info.block_id, info.block_number,
-                );
-                Some(info.block_number)
-            });
+        let epoch_change_block_number = last_ledger_info.as_ref().and_then(|li| {
+            let info = li.ledger_info().commit_info().epoch_block_info()?;
+            info!(
+                "recover_blocks: epoch change detected at block_id={}, block_number={}",
+                info.block_id, info.block_number,
+            );
+            Some(info.block_number)
+        });
 
         certs.sort_unstable_by_key(|qc| qc.commit_info().round());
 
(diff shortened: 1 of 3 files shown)


ByteYue (approved):

ByteYue (dismissed):

ByteYue:
P1: `send_for_execution(recovery)` needs to handle the case where suffix filtering leaves `blocks_to_commit` empty.

The current flow builds the original path, calls `init_block_number`, then filters out blocks whose `block_number > recover_epoch_change_block_number` in the recovery branch. However, it still unconditionally calls `commit_callback(&blocks_to_commit, ...)` afterwards. `BlockTree::commit_callback` immediately does `blocks_to_commit.last().expect("pipeline is empty")`, so a suffix-only replay proof can still panic during recovery.

One concrete scenario: the epoch-change block has already been replayed/committed by an earlier proof, and a later QC still carries `epoch_block_info` for a suffix block. At that point `path_from_ordered_root` can contain only suffix blocks; after filtering, `blocks_to_commit` is empty, and the subsequent `commit_callback` hits `pipeline is empty`.

Suggested fix: after filtering, explicitly skip the remaining recovery commit/update logic when the filtered batch is empty, or stop replaying once recovery reaches the epoch-change boundary. It would also be useful to add a regression test for a suffix-only recovery proof so this restart panic does not come back.

ByteYue (approved):

PR #704, opened 2026-05-09
fix(consensus): harden fast-forward recovery pipeline

Lchangliang:
## Summary

This PR now covers the private-mainnet recovery / fast-forward-sync failure modes found during the 2055785 stall investigation.

Main changes:

- Align consensus recovery root with execution when execution is still at block 0.
  - If `latest_block_number == 0`, recover from genesis LI instead of a newer metadata DB LI.
  - Preserve the existing `is_last_block_of_prev_epoch` branch for epoch-boundary recovery.
- Harden fast-forward sync around non-blocking epoch changes.
  - Only return ledger infos for blocks that the returned QCs can actually commit.
  - Only send `EpochChangeProof` after the local commit root reaches the epoch-change boundary.
  - Reset validator rand/buffer pipeline before FFsync rebuild to avoid BlockStore/pipeline state divergence.
- Make rand manager ordered-block intake idempotent for stale/dequeued rounds.
  - Track the highest dequeued round and drop stale ordered blocks that are pushed again.
  - This covers the core-4 duplicate-send case where round 28843 was pushed again and blocked round 28844.
- Replay ordered path after `add_certs`.
  - After SyncInfo certs are inserted, if `highest_ordered_cert` is ahead of `ordered_root`, call `send_for_execution` once to replay the missing ordered path into rand/buffer.
  - This covers the richard-1 case where FFsync/rebuild had 28844..28851 in BlockStore but the pipeline never received them.
- Surface broken execution channel sends as errors instead of silently returning `Ok(())`.

## Incident Notes

Two different symptoms were observed:

- **core-4 duplicate send**: round 28843 / block `84fb855d` was sent to execution twice. The second push created a stale rand queue entry ahead of round 28844, so randomness for 28844 was decided but never reached BufferManager.
- **richard-1 missing replay**: FFsync pulled blocks/QCs through round 28851, but recovery only executed up to the latest committed LI at round 28843. The node then advanced current round from SyncInfo and treated old 28844/28845 proposals as stale, leaving the ordered path present in BlockStore but not replayed into the pipeline.

The fixes are intentionally split:

- reset before validator FFsync rebuild prevents stale pipeline state from surviving a BlockStore rebuild;
- rand queue stale-drop is the safety net for duplicate ordered-block pushes;
- add-certs ordered-path replay fills the gap when BlockStore already has a higher ordered cert than the local pipeline root.

## Validation

```bash
cargo +nightly fmt
RUSTFLAGS='--cfg tokio_unstable' cargo +nightly check -p 'path+file:///Users/lightman/repos/gravity-sdk/aptos-core/consensus#aptos-consensus@0.1.0'
RUSTFLAGS='--cfg tokio_unstable' cargo +nightly test -p 'path+file:///Users/lightman/repos/gravity-sdk/aptos-core/consensus#aptos-consensus@0.1.0' rand::rand_gen::block_queue --lib
```

All passed locally.

### DIFF
--- a/aptos-core/consensus/src/block_storage/sync_manager.rs
+++ b/aptos-core/consensus/src/block_storage/sync_manager.rs
@@ -208,7 +208,23 @@ impl BlockStore {
         if let Some(tc) = sync_info.highest_2chain_timeout_cert() {
             self.insert_2chain_timeout_certificate(Arc::new(tc.clone()))?;
         }
+        self.replay_ordered_path_if_needed().await?;
         BLOCK_SYNC_GAUGE.set_with(&[], 0);
+        Ok(())
+    }
+
+    async fn replay_ordered_path_if_needed(&self) -> anyhow::Result<()> {
+        let ordered_root_round = self.ordered_root().round();
+        let highest_ordered_cert = self.highest_ordered_cert().as_ref().clone();
+        let highest_ordered_round = highest_ordered_cert.commit_info().round();
+        if ordered_root_round < highest_ordered_round {
+            info!(
+                "[BlockStore] replay ordered path after adding certs: ordered_root_round={}, highest_ordered_round={}",
+                ordered_root_round,
+                highest_ordered_round,
+            );
+            self.send_for_execution(highest_ordered_cert, false, None).await?;
+        }
         Ok(())
     }
 
--- a/aptos-core/consensus/src/block_storage/sync_manager.rs
+++ b/aptos-core/consensus/src/block_storage/sync_manager.rs
@@ -83,6 +83,44 @@ pub enum NeedFetchResult {
 }
 
 impl BlockStore {
+    fn is_epoch_change_li_boundary_locally_committed(&self, li: &LedgerInfoWithSignatures) -> bool {
+        let commit_root = self.commit_root();
+        // For non-blocking epoch changes, the epoch-change LI may commit a suffix block after the
+        // actual epoch boundary. Recovery intentionally filters out those suffix blocks and only
+        // commits up to `epoch_block_info`, so checking against `li.commit_info()` would prevent us
+        // from sending the epoch-change proof and leave execution waiting for the first suffix
+        // block.
+        let epoch_block_info = li.commit_info().epoch_block_info();
+        let committed = if let Some(epoch_info) = epoch_block_info {
+            match commit_root.block().block_number() {
+                Some(root_number) if root_number > epoch_info.block_number => true,
+                Some(root_number) if root_number == epoch_info.block_number => {
+                    commit_root.id() == epoch_info.block_id
+                }
+                _ => false,
+            }
+        } else {
+            commit_root.id() == li.commit_info().id()
+        };
+
+        if !committed {
+            warn!(
+                "[FastForwardSync] skip epoch change proof because local commit root has not reached \
+                 epoch-change boundary: boundary_block_id={}, boundary_block_number={:?}, \
+                 local_root_id={}, local_root_block_number={:?}, local_root_round={}",
+                epoch_block_info
+                    .map(|info| info.block_id)
+                    .unwrap_or_else(|| li.commit_info().id()),
+                epoch_block_info.map(|info| info.block_number),
+                commit_root.id(),
+                commit_root.block().block_number(),
+                commit_root.round(),
+            );
+        }
+
+        committed
+    }
+
     /// Check if we're far away from this ledger info and need to sync.
     /// This ensures that the block referred by the ledger info is not in buffer manager.
     pub fn need_sync_for_ledger_info(&self, li: &LedgerInfoWithSignatures) -> bool {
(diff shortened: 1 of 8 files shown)


ByteYue (commented):
Tracing the failure path on a stuck validator (fresh PD, started ~28s after the rest of the network was already at block 5+), this fix does land on the correct seam: `apply_fast_forward_sync` / `fast_forward_sync` validator branch calls `self.storage.start(...)` again after `update_latest_ledger_info()`, so the freshly-written block_3 LI in `metadata_db` was being read back through `aptos_db.get_latest_ledger_info()` and `LedgerRecoveryData::find_root` was anchoring `commit_root` at block 3 while reth was still at 0. With this patch the root stays at genesis, `recover_blocks` no longer skips blocks 1/2/3, and `send_for_execution` walks the chain forward, calling `set_ordered_blocks` per block — which is exactly the path that was missing for reth to catch up. So the fix is real, not a workaround.

A few comments on top of that:

1. **Doesn't cover `latest_block_number > 0` divergence.** The same root-vs-EL drift can happen mid-sync: e.g. a validator commits block 1 to reth, crashes, restarts with `metadata_db` already holding LI for block 5 from a previous `fast_forward_sync` that didn't get fed to EL. `latest_block_number == 0` won't fire and we fall through to `aptos_db.get_latest_ledger_info()` — same divergence, same stuck state. Consider generalizing to "look up the LI corresponding to `latest_block_number` from EL" (e.g. via `LedgerInfoSchema` keyed by block_number) and only fall back to genesis when that lookup fails. Could be a follow-up PR; flagging here so it's not forgotten.

2. **Test coverage.** The trigger conditions are very specific (fresh PD + late start + network commit ≥ 1 block by the time the node receives its first VoteMsg). A unit / simulation test for `StorageWriteProxy::start` would prevent regressions: mock `latest_commit_block_number()` returning 0, seed `aptos_db` with a non-genesis latest LI, assert the returned `RecoveryData` root is genesis.

3. **The `Ok(None) | Err(_)` split is a separate concern.** Splitting the match arm to log the underlying error is a clean improvement, but it's logically independent from the genesis-pinning fix. Either pull it into its own PR or add a one-liner in the description so reviewers don't have to figure out it's an unrelated cleanup.

4. **Branch ordering.** Putting the `latest_block_number == 0` check before `is_last_block_of_prev_epoch` is correct (block 0 = genesis can't simultaneously be "last block of previous epoch"), but a one-line comment stating that invariant would save the next maintainer a few minutes.

5. **Add an invariant log line.** At the end of `start()`, log `(execution_latest_block_number, recovery_root_round, recovery_root_block_number)` together. If this class of root/EL divergence ever recurs, a single grep against the consensus log will identify it immediately rather than requiring a metadata DB inspection.

Note for ops: this fix only takes effect at process startup. A node that's already stuck (e.g. core-4) won't self-heal after the binary lands — it still needs a wipe + restart so that the new `start()` path runs against a fresh PD.

ByteYue (dismissed):
Approving — fix is correct for the fresh-PD / late-start scenario we hit on core-4. The five points in my earlier comment are non-blocking; the EL>0 generalization (#1) can land as a follow-up PR.

Lchangliang:
Thanks for the detailed review.

On point 1, I agree this should be split into two cases:

- For the epoch-boundary case where `latest_block_number > 0` is the last block of the previous epoch, the existing `is_last_block_of_prev_epoch` branch is intentionally preserved. In that path we still read `LedgerInfoSchema(latest_block_number)` and use the LI that corresponds to the execution layer height, instead of blindly using `aptos_db.get_latest_ledger_info()`.
- The current patch only adds the missing special case for `latest_block_number == 0`, where the only correct recovery root is genesis. This is the concrete stuck-core-4 / fresh-PD failure mode we observed.

I intentionally did not generalize all non-zero mid-epoch recovery to always use `LedgerInfoSchema(latest_block_number)` in this PR, because that would change the previous recovery semantics for the normal non-epoch path. If we want to cover arbitrary mid-sync EL/metadata divergence (e.g. EL at block 1 while metadata latest LI is block 5), I agree that should be a follow-up with a focused test.

For point 3, agreed that the `Ok(None)` / `Err(e)` split is only logging cleanup. I can either mention it explicitly in the PR description or move it out if we want this patch to stay strictly minimal.

For point 4/5, both make sense. I can add the branch-order invariant comment and a recovery invariant log line in this PR if preferred.

ByteYue (dismissed):

Lchangliang:
Updated the PR title/body to match the current scope after the additional private-mainnet investigation.

The PR is no longer only the recovery-root alignment fix. It now covers three related recovery/FFsync hardening pieces:

- recovery root alignment when execution is still at block 0;
- stale rand queue protection for the core-4 duplicate ordered-block push;
- ordered-path replay after `add_certs` for the richard-1 case where BlockStore had a higher ordered cert than the local pipeline root.

The existing `is_last_block_of_prev_epoch` behavior is still preserved; the new genesis path is only for `latest_block_number == 0`.

ByteYue (dismissed):

ByteYue (approved):

PR #706, opened 2026-05-11
fix(cluster): default genesisTimestampSecs so e2e survives genesis-tool strict mode

ByteYue:
## Description

Upstream `gravity_chain_core_contracts` [PR #90](https://github.com/Galxe/gravity_chain_core_contracts/pull/90) makes `genesisTimestampSecs` **required** in the genesis-tool config (panics if absent) so that multi-operator genesis ceremonies produce byte-identical `genesis.json`. Before that PR, the genesis-tool silently fell back to the host's `SystemTime::now()`, which produced different EL storage between operators (most visibly the `OracleTask.updatedAt` slot under `OracleTaskConfig`) and broke the ceremony's sha256 reproducibility check.

This PR is the **companion change on the gravity-sdk side**, landed first so that nothing breaks the moment the `external/gravity_chain_core_contracts` ref is bumped.

## What this changes

`cluster/utils/aggregate_genesis.py` already had passthrough plumbing (`if "genesis_timestamp_secs" in genesis_cfg: …`), but **no fallback** — so any TOML that didn't set the field produced JSON without `genesisTimestampSecs`. Today that is fine (genesis-tool falls back to host clock); after the upstream merges, it would panic every e2e suite.

Two surgical edits:

1. `get_genesis_defaults()` now includes `"genesisTimestampSecs": 1778457600` (= **2026-05-11 00:00:00 UTC**), with a comment explaining why and pointing at the override path.
2. `build_genesis_config()` replaces the conditional `if … in genesis_cfg` block with the standard `genesis_cfg.get("genesis_timestamp_secs", defaults["genesisTimestampSecs"])` pattern used by every other field in the function.

## Why a fallback default rather than per-TOML edits

There are **18** `genesis.toml` files under `cluster/` and `gravity_e2e/cluster_test_cases/`; none of them set `genesis_timestamp_secs`:

```
$ grep -l genesis_timestamp_secs cluster/genesis.toml \
      cluster/example/*/genesis.toml \
      gravity_e2e/cluster_test_cases/*/genesis.toml
(no matches)
```

A single one-line default keeps all 18 suites booting with zero per-file edits. Suites that *do* care about reproducibility — ceremony rehearsal, future drift-regression tests — can still override `genesis.genesis_timestamp_secs` in their TOML and it wins via the standard `get(...)` pattern.

The chosen constant (1778457600 / 2026-05-11 00:00:00 UTC) is arbitrary; it just has to be deterministic and stable across machines. Picking a fixed point in time also makes any future log line that surfaces "updatedAt = …" obvious-at-a-glance rather than looking like a noise value.

## How Has This Been Tested?

Smoke-tested both branches of the new code path:

```python
$ python3 -c "
import sys; sys.path.insert(0, 'cluster/utils')
from aggregate_genesis import build_genesis_config, get_genesis_defaults

# fallback path (no override in genesis.toml)
r1 = build_genesis_config({}, {})
assert r1['genesisTimestampSecs'] == 1778457600

# override path (suite sets it in genesis.toml)
r2 = build_genesis_config({}, {'genesis_timestamp_secs': 1775664000})
assert r2['genesisTimestampSecs'] == 1775664000

# defaults dict exposes the key
assert 'genesisTimestampSecs' in get_genesis_defaults()
print('OK')
"
  fallback path: genesisTimestampSecs = 1778457600 ✓
  override path: genesisTimestampSecs = 1775664000 ✓
  defaults['genesisTimestampSecs'] = 1778457600 ✓
OK
```

The upstream-side verification (that the configured value actually lands in the `OracleTaskConfig.updatedAt` slot, and that two runs separated by host wall-clock produce byte-identical `genesis_accounts.json`) is documented in `gravity_chain_core_contracts` PR #90.

## Key Areas to Review

- The fallback constant value — happy to swap it for whatever the team prefers (e.g. a per-environment value, a build-time injection, or a different fixed timestamp). The value itself is not load-bearing; only its *stability* matters.
- The implicit policy change: e2e suites that don't override now share a fixed creation timestamp. If any existing assertion depends on `updatedAt > 0` *and the clock being recent*, it would silently be looking at 2026-05-11. I greped for `updatedAt` in `gravity_e2e/` and found no such assertions, but worth a second pair of eyes.

## Coordination with `gravity_chain_core_contracts`

- Upstream PR: https://github.com/Galxe/gravity_chain_core_contracts/pull/90
- Recommended merge order: this PR → upstream PR → bump `external/gravity_chain_core_contracts` ref in `cluster/genesis.toml`.
- If the order slips and upstream lands first, every e2e in this repo will panic at genesis generation with a clear, actionable message naming the missing field — no silent breakage, just CI red.

## Type of Change

- [x] Bug fix

## Which Components or Systems Does This Change Impact?

- [x] E2E Tests
- [x] CLI Tools  (`cluster/utils/aggregate_genesis.py` is invoked by the cluster bootstrap)

## Checklist

- [x] Self-reviewed the diff (15 +/3 - in a single file).
- [x] Comments added explaining the why and the override path.
- [x] Smoke-tested both fallback and override paths.
- [x] No new warnings; no behavior change for any suite that didn't already rely on the field being absent.

### DIFF
--- a/cluster/utils/aggregate_genesis.py
+++ b/cluster/utils/aggregate_genesis.py
@@ -34,6 +34,15 @@ def get_genesis_defaults():
         "consensusConfig": "0x0301010a00000000000000280000000000000001010000000a000000000000000100010200000000000000000020000000000000",
         "executionConfig": "0x00",
         "initialLockedUntilMicros": 1798848000000000,
+        # genesis-tool requires a deterministic EVM block.timestamp during
+        # genesis simulation so that independent operators produce byte-
+        # identical genesis.json (multi-operator ceremony reproducibility
+        # check). The genesis-tool panics if the field is absent; this
+        # default keeps every e2e suite booting without per-TOML edits.
+        # Ceremonies / suites that need a specific value should override
+        # `genesis.genesis_timestamp_secs` in their genesis.toml.
+        # 1778457600 = 2026-05-11 00:00:00 UTC.
+        "genesisTimestampSecs": 1778457600,
         "validatorConfig": {
             "minimumBond": "1000000000000000000",
             "maximumBond": "1000000000000000000000000",
@@ -99,9 +108,12 @@ def build_genesis_config(config, genesis_cfg):
     result["executionConfig"] = genesis_cfg.get("execution_config", defaults["executionConfig"])
     result["initialLockedUntilMicros"] = genesis_cfg.get("initial_locked_until_micros", defaults["initialLockedUntilMicros"])
 
-    # Optional: genesis block timestamp (passthrough only if explicitly set)
-    if "genesis_timestamp_secs" in genesis_cfg:
-        result["genesisTimestampSecs"] = genesis_cfg["genesis_timestamp_secs"]
+    # Genesis EVM block.timestamp — required by genesis-tool for reproducible
+    # genesis.json output. Always emitted; falls back to the deterministic
+    # default constant when not set per-suite in genesis.toml.
+    result["genesisTimestampSecs"] = genesis_cfg.get(
+        "genesis_timestamp_secs", defaults["genesisTimestampSecs"]
+    )
 
     # governanceOwner is now a required field in genesis-tool's GenesisConfig
     # (contracts main commit 57ae9bc wires Governance.owner at genesis via


Lchangliang (approved):

nekomoto911 (approved):

Richard1048576 (approved):
LGTM

Richard1048576 (approved):
LGTM

PR #722, opened 2026-05-19
mempool: per-entry TTL + bucket-sharded snapshot + silent-blackhole recovery

nekomoto911:
## Summary

Silent black-hole recovery for the mempool broadcast path — and adds an e2e regression that **deterministically** exercises the recovery code path.

Three commits:

1. **`feat(mempool):`** Replace global-wipe `TxnCache` with per-entry TTL + bucket-sharded snapshot + four-state filter (`Dispatch / WaitForPrimary / SuppressInTtl / SuppressSameSlot`). A tx whose first dispatch landed on a stuck Primary slot is now auto-routed via the Failover slot one TTL (default 5s) later — no gaptos changes, no global re-broadcast, no extra tokio tasks.

2. **`test(e2e):`** Add Phase 3 to `pfn_chain` e2e. Restarts pfn1 with `GRAVITY_BLACKHOLE_BROADCAST=1` (Mempool-side debug env knob, ~10 LOC in `bin/gravity_node/src/mempool.rs`) so pfn1 stays a healthy member of `sync_states` but silently drops mempool broadcasts; then drives 30s of multi-account load via pfn3 and asserts impl-d's slot-flip catches every in-flight tx within ~1 TTL + commit.

3. **`test(e2e):`** Refactor Phase 3 into **two back-to-back halves** — Half A blackholes pfn1, Half B blackholes pfn2. pfn3 is never restarted between halves so `priority.rs`'s `RandomState`-seeded Primary stays fixed → exactly one half has `Primary == blackhole target` and exercises slot-flip, the other half hits direct delivery latency. The cross-half p95 split is a **deterministic** assertion — no more `--force-init` coverage lottery.

## Phase 3 assertions

Per-half:
- `sent ≥ 150`, `timeout == 0`, `failed == 0`, `p99 ≤ 24s` (sanity ceiling)

Cross-half (deterministic bimodal):
- `min(p95) ≤ 3.0s` — one half must run at direct delivery latency
- `max(p95) ≥ 6.0s` — other half must hit slot-flip (TTL=5s + commit ~1s)
- `gap ≥ 4.0s` — separation between the two modes

## Test plan

- [x] First clean run on `mempool-impl-d` worktree (no `--force-init`):
  ```
  Half A (pfn1 blackhole): 210/0/0  p50=6.25  p95=8.61  p99=8.93
  Half B (pfn2 blackhole): 233/0/0  p50=1.18  p95=2.05  p99=2.48
  bimodal: fast=2.05  slow=8.61  gap=6.57   PASS
  Total wall time: 6:16
  ```
  Half A p50 = 6.25s ≈ TTL(5s) + commit(~1s) — slot-flip exactly at the predicted floor.
- [ ] Reviewer: run `SKIP_CONTRACTS_FETCH=1 python3 gravity_e2e/runner.py pfn_chain` on a fresh worktree; expect either Half A or Half B to land in the slow mode (Primary assignment is RandomState-dependent across machines).
- [ ] Phase 1 / Phase 2 still pass unmodified.
- [ ] Verify `GRAVITY_BLACKHOLE_BROADCAST` env knob has no effect when unset (default node behaviour unchanged).

### DIFF
--- a/aptos-core/mempool/src/core_mempool/mempool.rs
+++ b/aptos-core/mempool/src/core_mempool/mempool.rs
@@ -44,8 +44,15 @@ pub struct TxnCache {
 
 #[derive(Clone, Copy)]
 struct CacheEntry {
+    /// For dispatched entries: when the tx was last handed to a peer.
+    /// For placeholders (`dispatched == false`): the Failover first-sighting
+    /// time — i.e. when the TTL grace clock started ticking.
     last_dispatched_at: Instant,
     last_target: TargetSlot,
+    /// `false` ⇒ placeholder seeded by a Failover first-sighting awaiting
+    /// Primary claim within `cache.ttl`. `true` ⇒ tx has been dispatched
+    /// at least once (the normal in-TTL-suppress / TTL-re-emit regime).
+    dispatched: bool,
 }
 
 /// `(bucket, priority_discriminant)` — a zero-cost proxy for the destination
@@ -197,32 +204,43 @@ impl CoreMempoolTrait for Mempool {
             if out.len() >= count {
                 break;
             }
+            // PR #722 review point 3: the TTL cache is now self-sufficient
+            // for failover semantics. Primary first-sighting dispatches
+            // immediately. Failover first-sighting seeds a placeholder so
+            // the TTL clock starts here. Within the `cache.ttl` grace,
+            // Primary can still claim the placeholder (preserves the
+            // Primary-first invariant). After the grace elapses, Failover
+            // takes over — no dependency on `priority.rs` promotion.
             let dispatch = match cache.entries.get(&entry.hash) {
-                // First sighting of this hash: only Primary may take it.
-                // Holding Failover here keeps the upstream invariant
-                // "Primary gets first shot at every tx, Failover only
-                // catches misses" — without this, the slot whose tick
-                // happens to land first wins, which lets a Failover
-                // ahead of Primary steal first-dispatch.
-                //
-                // If the Primary peer is unhealthy and never queries,
-                // `priority.rs::update_peers_interval` (default 1s)
-                // promotes the surviving peer to Primary; the next
-                // read_timeline from that peer arrives with
-                // `priority_of_receiver = Primary` and dispatches normally.
-                None if matches!(priority_of_receiver, BroadcastPeerPriority::Primary) => true,
-                None => false,
+                None => matches!(priority_of_receiver, BroadcastPeerPriority::Primary),
+                Some(e) if !e.dispatched => match priority_of_receiver {
+                    BroadcastPeerPriority::Primary => true,
+                    BroadcastPeerPriority::Failover => {
+                        now.duration_since(e.last_dispatched_at) >= cache.ttl
+                    }
+                },
                 Some(e) if now.duration_since(e.last_dispatched_at) < cache.ttl => false,
                 Some(e) if e.last_target == target_slot && priority_count >= 2 => false,
                 Some(_) => true,
             };
             if !dispatch {
+                // Failover first-sighting seeds a placeholder so the TTL
+                // clock starts. `or_insert` (not `insert`) preserves the
+                // original first_seen_at across repeated Failover ticks
+                // during the grace window.
+                if matches!(priority_of_receiver, BroadcastPeerPriority::Failover) {
+                    cache.entries.entry(entry.hash).or_insert(CacheEntry {
+                        last_dispatched_at: now,
+                        last_target: target_slot,
+                        dispatched: false,
+                    });
+                }
                 continue;
             }
             out.push((entry.txn, 0));
             cache.entries.insert(
                 entry.hash,
-                CacheEntry { last_dispatched_at: now, last_target: target_slot },
+                CacheEntry { last_dispatched_at: now, last_target: target_slot, dispatched: true },
             );
         }
         let len = out.len();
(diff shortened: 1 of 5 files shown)


keanji-x:
Reviewing this as a code/test stability issue, I see a few blockers:

1. `gravity_e2e/cluster_test_cases/pfn_chain/test_pfn_chain.py:900` / `:903` / `:937`

   The Phase 3 assertion assumes that because `pfn3` is not restarted, the same upstream remains Primary across both halves, so exactly one half must be fast and one half must be slow. That does not hold with the current gaptos priority code: peer ordering can change from peer monitoring/ping data and from the stop/start of `pfn1` / `pfn2`, not just from the `RandomState`. This makes the p95 bimodal check flaky: CI can get both halves fast, both halves slow, or a middle distribution even when the mempool implementation is behaving correctly.

   I think this test should either observe/fix the actual Primary before blackholing it, or assert coverage using the real priority assignment/logs instead of inferring it from latency.

2. `aptos-core/mempool/src/core_mempool/mempool.rs:51` / `:216`

   `TargetSlot = (bucket, priority)` is not a stable peer identity. The same `(bucket, Primary)` can point to a different peer after priority reassignment, and `(bucket, Failover)` can also change underneath us. The cache then suppresses/permits retries based on a logical priority label rather than the real destination peer, so it can suppress a retry that should go to a newly selected peer, or treat an alternate slot as different when it is not a different destination.

   This should track the actual `PeerNetworkId` for the dispatch target, or at least include a priority-assignment generation in the target key.

3. `aptos-core/mempool/src/core_mempool/mempool.rs:213`

   New transactions are only dispatched on first sighting if the caller is `Primary`; a Failover caller returns nothing and does not create a cache entry. That means the TTL/degrade logic never starts for a transaction if Primary is temporarily not querying but Failover is. The comment says the surviving peer will be promoted to Primary, but that is a separate mechanism and not guaranteed within this cache's TTL semantics.

   Please add coverage for a failover-only / primary-not-querying case, or change the first-dispatch path so an active Failover can eventually take ownership after the intended delay.

nekomoto911:
Concur the structural concern is real — `(bucket, priority)` is a point-in-time mapping, so across a `priority.rs` reassignment the same slot label may resolve to a different peer.

**Practical impact is narrow.** The misjudgment window is exactly: a reassignment that lands inside one `TTL=5s` interval *and* swaps which peer owns a `(bucket, priority)` slot. With a stable peer set the reassignment does not happen; the silent-blackhole path this PR targets keeps the peer set fixed. Worst case inside the window is one duplicate or one delayed send of a single tx, recovered on the next TTL cycle — no livelock, since TTL rotation breaks any wrong decision automatically.

**Cost to fix properly is non-trivial.** Threading real peer identity requires `peer: PeerNetworkId` in `CoreMempoolTrait::read_timeline`. The trait lives in gaptos (`shared_mempool/types.rs:64`); the call site at `network.rs:526` already has `peer` in scope, so the gaptos patch is ~5 LOC mechanically — but it requires forking gaptos and pinning a paired `rev` bump, which is exactly the constraint this PR set out to avoid. A generation counter inside `TargetSlot` does not work locally either: reassignment originates in `prioritized_peers_state` on the gaptos side, and `read_timeline` carries no peer-bearing channel, so Mempool can't observe the remap.

**Conclusion:** the trade-off does not pencil out for this PR. The slot proxy is correct in steady state, the failure window is narrow and self-healing within one TTL, and the only honest fix expands the touch surface in ways the design explicitly chose to avoid. I'd rather document the across-reassignment limitation in the `TargetSlot` doc-comment and leave the peer-id migration as a follow-up if real-world traces ever show it firing.

nekomoto911:
You're right that the bimodal assertion's stability rests on `compare_intelligent`'s ping_latency tiebreaker, which is real measured data and re-probed when pfn1/pfn2 stop+start between halves. Even on localhost, the assumption that `compare_intelligent`'s first three tiers (network_id, validator_distance, ping_latency) all tie isn't a strict invariant — peer monitoring, restart-driven re-handshake, or routine ping jitter can flip the Primary choice mid-run or across halves, and the bimodal pattern can collapse to two-fast, two-slow, or middle-distribution false fails even when impl-d is working correctly.

We're simplifying. The Phase 3 latency assertion now becomes a single SLA bound:

```python
P99_CEILING = 2 * MEMPOOL_TTL_SECONDS + 2.0  # = 12s
for half in (half_a, half_b):
    assert half["timeout"] == 0
    assert half["failed"] == 0
    assert half["p99"] <= P99_CEILING
```

`fast_p95` / `slow_p95` / `gap` and the entire bimodal block are gone.

Rationale: the bimodal assertion was implicitly a *coverage* assertion — forcing the test to hit the slot-flip code path by requiring one half to be slow. But branch coverage of slot-flip belongs in unit tests (`ttl_expired_alt_slot_dispatches` in `core_mempool/mempool.rs` exercises it deterministically), not in e2e via latency-shape inference. E2E should test system invariants under stress; the SLA bound (≤ 2 × TTL + commit slack) is exactly that — it's satisfied by both the direct dispatch (~1.5s) and slot-flip (~6.5s) paths.

Trade-off: if priority.rs happens to put the healthy peer as Primary in both halves, the slot-flip path isn't exercised by that run. We accept this false-negative window because the unit test covers the branch deterministically; the e2e's job is to catch system-level breakage (tx wedge, timeout, broadcast collapse), which `p99 <= 12s` does directly.

nekomoto911:
Correction to my earlier reply on point 3 — I mis-framed the scenario.

The reviewer asked about "Primary temporarily not querying but Failover is." My Cases A and C address different scenarios where Primary is still calling `read_timeline` (broadcast blackhole, backoff with `scheduled_backoff` retries). Case B concedes the architectural critique — "`priority.rs::update_peers_interval` ≤1s promotes the survivor" is exactly the cross-layer dependency the reviewer flagged — and defends it on empirical Phase 1b numbers rather than on layering.

The realistic failure mode my walkthrough misses: Primary peer alive (TCP healthy ⇒ `priority.rs` does not reassign) but its `read_timeline` call stalled — slow syscall, app-layer deadlock, broadcaster task starved. Cases A/B/C don't cover this, and `None + Failover -> no dispatch` does not seed a cache entry either, so the TTL clock never starts.

Pushed 58e8026c to take the second branch of your suggestion ("change the first-dispatch path so an active Failover can eventually take ownership after the intended delay"):

- `CacheEntry` gains `dispatched: bool`. Failover first-sighting now seeds a placeholder (`dispatched=false`) via `or_insert`, so the TTL clock starts immediately.
- Within the grace window Primary can still claim the placeholder anytime — Primary-first invariant preserved.
- After the grace elapses, Failover takes over without depending on `priority.rs` promotion.
- Grace window reuses `cache.ttl`, so the e2e SLA ceiling `2 × TTL + slack` bounds both direct and slot-flip paths consistently.

New unit tests cover the three transitions:

- `failover_first_sighting_creates_placeholder_no_dispatch`
- `primary_claims_placeholder_within_grace`
- `failover_takes_over_after_grace`

Doc-comment on the dispatch match cites this review point so future readers find the rationale.

keanji-x (dismissed):

keanji-x (approved):

PR #723, opened 2026-05-19
fix(consensus): pipeline / block-store fixes from Coacker audit triage

keanji-x:
Closes Galxe/gravity-audit#53
Closes Galxe/gravity-audit#54
Closes Galxe/gravity-audit#517
Closes Galxe/gravity-audit#522

## Summary

Four targeted fixes from triaging the Coacker audit backlog in `Galxe/gravity-audit`. Scope is intentionally narrow — only the pipeline / block-store findings whose fix didn't require a design discussion. The remaining audit findings (consensus protocol layer, leader-election randomness, upstream backports of aptos-labs PR #14637 / #14644) are tracked separately.

> Note on #517: this PR fixes only the externally-reachable panic (line 515 — missing parent block). The other three panic points cited in #517 (lines 518 / 731 / 896) are internal invariants where halting is the correct behavior; analysis posted as a comment on the issue. The `Closes #517` keyword applies because the issue is fully resolved by code-fix + analysis-comment combined.

## Fixes

| Audit issue | Commit | Severity | Change |
|---|---|---|---|
| Galxe/gravity-audit#53 | `1fd21af` | Critical | `buffer_item.rs`: use `commit_info.timestamp_usecs()` instead of `block.timestamp_usecs()` in `generate_executed_item_from_ordered` so the BlockInfo stays consistent across epoch boundaries (where `change_timestamp()` has been applied). Without this, the suffix-block path can deterministically panic at the later commit-proof assert. |
| Galxe/gravity-audit#54 | `7559cc1` | Critical | `buffer_manager.rs`: persisting-phase error arm previously only logged and continued, letting the pipeline silently stall after committed blocks had been popped. Convert to an explicit panic so the supervisor restarts. (Full in-process reset is still a TODO, but loud crash > silent stall.) |
| Galxe/gravity-audit#517 | `f059b24` | Critical | `block_store.rs`: convert the "cannot find parent_block id" panic in `init_block_number` to `anyhow::bail!` and propagate from `send_for_execution`. The other three panic points in this file flagged by the audit (line 518 / 731 / 896) are intentionally kept — those are real invariant violations where panicking is correct. |
| Galxe/gravity-audit#522 | `b08c94e` | High | `block_buffer_manager.rs`: remove the redundant unlocked `is_epoch_change()` early check in `get_ordered_blocks`. The in-lock recheck at the start of the wait loop already handles the same condition, so the pre-check only widened the TOCTOU window without changing semantics. |

## Verification

- `RUSTFLAGS="--cfg tokio_unstable" cargo check --manifest-path aptos-core/consensus/Cargo.toml` ✓
- `RUSTFLAGS="--cfg tokio_unstable" cargo check --manifest-path crates/block-buffer-manager/Cargo.toml` ✓

## Out of scope (tracked separately)

- Galxe/gravity-audit#41 + #43 — backport aptos-labs/aptos-core#14637 ("Sync up QC in order vote message")
- Galxe/gravity-audit#67 — backport aptos-labs/aptos-core#14644 (refactor `IncrementalProofState::take` → `aggregate_and_verify` returning Result)
- Galxe/gravity-audit#45 — `initiate_new_epoch` shutdown-before-sync ordering (upstream has the same bug; needs design)
- Galxe/gravity-audit#518 — analysis showed all 5 cited panic points are caller-guarded; recommend closing the audit issue with that analysis rather than a code change
- Galxe/gravity-audit#523 — `leader_reputation` `root_hash = HashValue::zero()`; needs design discussion on randomness source (VRF vs. Reth-derived hash)

## Test plan

- [ ] CI green
- [ ] Existing consensus tests pass under `cargo test`
- [ ] Spot-check that the #53 fix clears the assert in epoch-transition path (would benefit from a unit test; not added in this PR)
- [ ] Spot-check that the #522 fix doesn't regress epoch-change response latency (semantically same; only one fewer unlocked atomic read on the fast path)

### DIFF
--- a/aptos-core/consensus/src/block_storage/block_store.rs
+++ b/aptos-core/consensus/src/block_storage/block_store.rs
@@ -501,7 +501,10 @@ impl BlockStore {
         block_store
     }
 
-    pub fn init_block_number(&self, ordered_blocks: &Vec<Arc<PipelinedBlock>>) {
+    pub fn init_block_number(
+        &self,
+        ordered_blocks: &Vec<Arc<PipelinedBlock>>,
+    ) -> anyhow::Result<()> {
         let mut block_numbers = vec![];
         let mut block_number = 0;
         for p_block in ordered_blocks {
@@ -512,7 +515,17 @@ impl BlockStore {
             {
                 block_number = parent_block.block_number().unwrap() + 1;
             } else {
-                panic!("Cannot find the parent_block id {}", p_block.parent_id());
+                // The parent block is reachable from neither the in-memory tree nor
+                // persistent storage. This can happen when an externally-sourced ordered
+                // block references an unknown ancestor (e.g. malformed sync data on
+                // startup). Surface as a recoverable error so the caller can log/abort
+                // gracefully instead of crashing the node mid-pipeline.
+                anyhow::bail!(
+                    "init_block_number: cannot find parent block id {} (epoch {}, block {})",
+                    p_block.parent_id(),
+                    p_block.epoch(),
+                    p_block.block().id(),
+                );
             }
             if let Some(cur_block_number) = p_block.block().block_number() {
                 assert_eq!(cur_block_number, block_number);
@@ -529,6 +542,7 @@ impl BlockStore {
         if block_numbers.len() != 0 {
             self.storage.save_tree(vec![], vec![], block_numbers).unwrap();
         }
+        Ok(())
     }
 
     /// Send an ordered block id with the proof for execution, returns () on success or error
@@ -589,7 +603,7 @@ impl BlockStore {
         self.pending_blocks.lock().gc(finality_proof.commit_info().round());
         // This callback is invoked synchronously with and could be used for multiple batches of
         // blocks.
-        self.init_block_number(&blocks_to_commit);
+        self.init_block_number(&blocks_to_commit)?;
         if recovery {
             // Recovery mode: process blocks directly without going through execution pipeline.
             // Filter out suffix blocks past the epoch change boundary before execution.
--- a/crates/block-buffer-manager/src/block_buffer_manager.rs
+++ b/crates/block-buffer-manager/src/block_buffer_manager.rs
@@ -605,9 +605,11 @@ impl BlockBufferManager {
             self.ready_notifier.notified().await;
         }
 
-        if self.is_epoch_change() {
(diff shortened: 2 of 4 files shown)


crates/block-buffer-manager/src/block_buffer_manager.rs
```diff
@@ -605,9 +605,11 @@ impl BlockBufferManager {
             self.ready_notifier.notified().await;
         }
 
-        if self.is_epoch_change() {
-            return Err(anyhow::anyhow!("Buffer is in epoch change"));
-        }
+        // Note: the previous early `is_epoch_change()` check here was redundant
```
keanji-x, line 608: why

aptos-core/consensus/src/block_storage/block_store.rs
```diff
@@ -589,7 +603,7 @@ impl BlockStore {
         self.pending_blocks.lock().gc(finality_proof.commit_info().round());
         // This callback is invoked synchronously with and could be used for multiple batches of
         // blocks.
-        self.init_block_number(&blocks_to_commit);
+        self.init_block_number(&blocks_to_commit)?;
```
keanji-x, line 606: why

keanji-x:
Closing after independent re-review; superseded by #773 + follow-ups. Disposition of the four fixes:

- **#53 (timestamp)** — already fixed on `main` by #735 via the hardfork-gated `use_corrected_commit_timestamp` (`ConsensusHardfork::ConsensusAlpha`). That gating is the correct approach: this PR's unconditional version would change the signed `BlockInfo` and split consensus with un-upgraded nodes, so it must not be merged as-is.
- **#54 (persist-failure panic)** — reviewed correct but incomplete: `ExecutionProxy::commit` structurally never returns `Err` (all failure paths are already `.expect()`), so the new panic arm is dead code. The real silent-stall path is the persisting-phase task dying and `persisting_phase_rx` yielding `None`, which the `select!`'s `Some(result)` pattern silently disables. Will be re-raised handling the `None` case.
- **#517 (panic → bail)** — reviewed correct (all callers handle the Result), but it trades fail-fast for fail-soft: a persistently missing parent now leaves the node alive-but-stalled instead of crash-restarting. Deferred to next release together with an alert on the new `cannot find parent block` log line.
- **#522 (drop unlocked pre-check)** — reviewed correct, no urgency; will ride along with the re-raised PR.

#54/#517/#522 will come back as a fresh PR rebased on current main.

PR #726, opened 2026-05-20
fix(consensus): return Result from IncrementalProofState::take instead of panicking

keanji-x:
## Summary
- Minimal surgical fix for [gravity-audit#67](https://github.com/Galxe/gravity-audit/issues/67) (Critical).
- Replaces `unreachable!()` in `IncrementalProofState::take` with a real `Result` return so the `ProofCoordinator` task survives a BLS aggregation failure (e.g. after validator-set rotation) instead of crashing and halting consensus.
- Adds `SignedBatchInfoError::UnableToAggregate` and propagates `?` at the single caller.

## Why minimal (not a full backport of upstream #14644)
Upstream aptos-core PR [#14644](https://github.com/aptos-labs/aptos-core/pull/14644) fixes the same panic but as part of a much larger refactor (~466 lines of conflicts against our vendored consensus). This PR ports only the panic→Result slice; we can decide later whether to take the rest of the refactor.

## Test plan
- [x] `cargo check -p aptos-consensus --tests` passes.
- [ ] Reviewer: confirm no other in-tree callers of `IncrementalProofState::take` (there is only one, in `aggregate_and_verify`).

### DIFF
--- a/aptos-core/consensus/consensus-types/src/proof_of_store.rs
+++ b/aptos-core/consensus/consensus-types/src/proof_of_store.rs
@@ -274,6 +274,7 @@ pub enum SignedBatchInfoError {
     NotFound,
     AlreadyCommitted,
     NoTimeStamps,
+    UnableToAggregate,
 }
 
 #[derive(Deserialize, Serialize, Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


keanji-x:
Superseded by #732 (combined PR with the #41+#43 fixes from #727; build + clippy + targeted test verified locally).

PR #727, opened 2026-05-20
fix(consensus): equivocation detection + round-jump guard for order votes

keanji-x:
## Summary
Minimal surgical fixes for two related gravity-audit findings in the order-vote path:

- **[#41](https://github.com/Galxe/gravity-audit/issues/41) (Critical)** — `PendingOrderVotes` now tracks `author -> LedgerInfo digest` and detects equivocation, mirroring `PendingVotes::insert_vote`. A Byzantine validator can no longer split its voting power across conflicting ordering candidates.
- **[#43](https://github.com/Galxe/gravity-audit/issues/43) (High)** — `process_order_vote_msg` now snapshots `highest_ordered_round` before the embedded QC is inserted and rejects votes whose round jumps more than `MAX_ORDER_VOTE_ROUND_JUMP` past that snapshot. Closes the fast-forward window where a single valid QC could be used to land an order vote for a round we never sequentially processed.

## Why minimal (not a full backport of upstream #14637)
Upstream aptos-core PR [#14637](https://github.com/aptos-labs/aptos-core/pull/14637) covers both findings, but does so by reshaping `li_digest_to_votes` into a `(QuorumCert, OrderVoteStatus)` map and plumbing `verified_quorum_cert: Option<QuorumCert>` through `insert_order_vote`. That refactor conflicts on ~164 lines in our vendored consensus and changes call sites we'd then have to re-audit. This PR keeps the original types and only adds the security checks.

## Test plan
- [x] `cargo check -p aptos-consensus --tests` passes.
- [ ] Reviewer: confirm `MAX_ORDER_VOTE_ROUND_JUMP` value is safe vs. legitimate validator sync flows.
- [ ] Reviewer: confirm equivocation log uses the same `SecurityEvent::ConsensusEquivocatingVote` channel as `PendingVotes`.

### DIFF
--- a/aptos-core/consensus/src/pending_order_votes.rs
+++ b/aptos-core/consensus/src/pending_order_votes.rs
@@ -21,6 +21,8 @@ pub enum OrderVoteReceptionResult {
     /// The vote has been added but QC has not been formed yet. Return the amount of voting power
     /// QC currently has.
     VoteAdded(u128),
+    /// The very same author has already voted for another LedgerInfo digest (equivocation).
+    EquivocateVote,
     /// This block has just been certified after adding the vote.
     NewLedgerInfoWithSignatures(LedgerInfoWithSignatures),
     /// There might be some issues adding a vote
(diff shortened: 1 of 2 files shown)


keanji-x:
Superseded by #732 (combined PR with the #67 fix from #726; build + clippy + targeted test verified locally).

PR #741, opened 2026-06-10
test(e2e): regression for self-sponsored EIP-7702 chain halt (audit #677)

keanji-x:
## What

Adds a single-node e2e regression suite (`gravity_e2e/cluster_test_cases/prague_7702_selfsponsor/`) for **gravity-audit #677** — a CRITICAL, externally-triggerable, deterministic chain halt.

## The bug

A **self-sponsored** EIP-7702 (type-4) transaction — sender is also the `authority` of an authorization tuple in its own `authorization_list` — bumps the sender's nonce **twice** during execution:
- once in `deduct_caller` (revm `pre_execution.rs:170`)
- once in `apply_eip7702_auth_list` (revm `pre_execution.rs:251`), because `authority == caller`

so the caller's post-state nonce is `expect + 2`. The **currently-pinned** grevm (`v2.2.4` / `26b586c`) hard-asserts `assert_eq!(change.info.nonce, expect + 1)` at `async_commit.rs:78` and **panics**.

The tx is spec-valid and pool-valid, so any funded EOA can submit it over public JSON-RPC; every validator then panics deterministically on the ordered block (and re-panics on replay) → **permanent network halt**. This is the bug that halted the testnet on 2026-06-09 (block 1400867 → 1400868).

The fix exists in a newer grevm rev (`>= 3c09e7c`, which allows a post-nonce in `[expect+1, expect+1+self_auth_count]`) but is **not pinned** by gravity-sdk, so the deployed chain remains exposed even though #677 was closed.

## Reproduced locally (gate evidence)

Against the pinned build, the self-sponsored 7702 tx kills the node — process exits, 14 RPC failures, zero block progress:

```
panicked at .../grevm/26b586c/src/async_commit.rs:78:29:
assertion `left == right` failed
  left: 2
 right: 1
```

→ test reports **XFAIL** (expected failure: bug present).

## Test design / why it's trustworthy

- **Prague coverage gap:** the shared devnet genesis doesn't set `pragueTime`, so a type-4 tx is rejected by the reth pool (`transaction type not supported`) **before** execution — the very gap that let #677 ship. `hooks.py:pre_start` enables Prague for this suite only.
- **Ordering fix:** `deploy.sh` copies the suite genesis to `<base_dir>/genesis.json` and boots the node from it *before* `pre_start` runs, so patching only the source artifact is too late and yields a Prague-disabled node + false-negative pass. The hook patches the **deployed** copy.
- **No false negatives:** the test **fails hard** (not xfail) if the tx is rejected pre-execution, and requires positive proof of execution (receipt present + sender nonce advances by exactly 2) before ever concluding the node is healthy. A Prague-disabled or tx-dropping harness can never masquerade as "fixed."
- **xfail (strict=False):** CI stays green until the grevm bump lands; it then flips to **xpass** to prompt removal of the marker.

## Scope

Test-only — adds one e2e suite, no production-code changes. Remove the `xfail` once grevm is bumped to `>= 3c09e7c` and relocked.

### DIFF
--- a/gravity_e2e/cluster_test_cases/prague_7702_selfsponsor/cluster.toml
+++ b/gravity_e2e/cluster_test_cases/prague_7702_selfsponsor/cluster.toml
@@ -1,0 +1,28 @@
+# Gravity Cluster Configuration - Prague EIP-7702 Self-Sponsor Suite
+#
+# Single validator node used to verify the node's handling of a self-sponsored
+# EIP-7702 (type-4) transaction (authority == sender). Requires Prague active —
+# enabled for THIS suite only via hooks.py:pre_start (no shared genesis change).
+# Distinct base_dir + ports from other suites so it never collides.
+
+[cluster]
+name = "gravity-devnet-7702"
+base_dir = "/tmp/gravity-cluster-7702"
+
+[genesis_source]
+genesis_path = "./artifacts/genesis.json"
+waypoint_path = "./artifacts/waypoint.txt"
+
+[[nodes]]
+id = "node1"
+role = "genesis"
+source = { project_path = "../" }
+host = "127.0.0.1"
+validator_port = 6482
+vfn_port = 6492
+rpc_port = 8845
+metrics_port = 9303
+inspection_port = 10302
+https_port = 1324
+authrpc_port = 8853
+reth_p2p_port = 12326
--- a/gravity_e2e/cluster_test_cases/prague_7702_selfsponsor/genesis.toml
+++ b/gravity_e2e/cluster_test_cases/prague_7702_selfsponsor/genesis.toml
@@ -1,0 +1,84 @@
+# Gravity Genesis Configuration - Prague EIP-7702 Self-Sponsor Suite
+#
+# Identical to the single_node suite (one genesis validator, devnet faucet),
+# only the validator_port / vfn_port differ so this cluster can run alongside
+# single_node without port collisions.
+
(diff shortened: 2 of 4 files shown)


keanji-x:
Closing as redundant — built on a stale local checkout.

After pulling latest main I confirmed:
- **#677 is already FIXED on main** by #738 (merged 2026-06-09), which bumped greth to `0adbb4c9` and grevm to `3c09e7c` (the rev that relaxes the `async_commit` nonce assertion to `[expect+1, expect+1+self_auth_count]`).
- The self-sponsored / authority==caller halt path is **already covered** by the Prague acceptance suite `gravity_e2e/cluster_test_cases/prague/test_eip7702.py` (case **P-B6**: "self-sponsored self-delegation (tx.from == authority) does not halt the chain"), added in #734/#738.

My suite only "reproduced" the panic because my checked-out `Cargo.lock` and prebuilt binary predated #738. There's no need for a duplicate xfail suite for an already-fixed, already-covered bug. Apologies for the noise — I should have pulled latest main and checked merged PRs before filing.

#740 (BLS PoP precompile gas-limit, audit #678) is unaffected: I re-verified that bug is still **unfixed** at the currently-pinned greth `0adbb4c` (`bls_precompile.rs:102` returns flat `POP_VERIFY_GAS` with no gas-limit guard; alloy-evm `precompiles.rs:425` still asserts on underflow).

PR #750, opened 2026-06-15
Gracefully stop reth coordinator tasks

ByteYue:
## Summary
- replace the coordinator watchdog's direct process::exit path with an awaited Result-returning shutdown path
- broadcast shutdown when a coordinator task exits, then wait briefly for the remaining coordinator tasks to stop
- join the reth runner thread before returning a non-zero exit for coordinator failures

## Validation
- RUSTFLAGS="--cfg tokio_unstable" cargo check -p gravity_node

Related audit context: Galxe/gravity-audit#11

### DIFF
--- a/bin/gravity_node/src/main.rs
+++ b/bin/gravity_node/src/main.rs
@@ -66,13 +66,13 @@ fn run_reth(
     cli: Cli<EthereumChainSpecParser>,
     execution_args_rx: oneshot::Receiver<ExecutionArgs>,
     mut shutdown: broadcast::Receiver<()>,
-) -> (ConsensusArgs<impl RethEthCall>, u64, oneshot::Receiver<PathBuf>) {
+) -> (ConsensusArgs<impl RethEthCall>, u64, oneshot::Receiver<PathBuf>, thread::JoinHandle<()>) {
     let (datadir_tx, datadir_rx) = oneshot::channel::<PathBuf>();
     reth_cli_util::sigsegv_handler::install();
 
     let (tx, rx) = std::sync::mpsc::sync_channel(1);
 
-    std::thread::spawn(move || {
+    let reth_thread = std::thread::spawn(move || {
         // Trick code to ensure the `rx.recv` won't panic before the error message is printed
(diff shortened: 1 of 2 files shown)


nekomoto911:
Review finding, worth verifying before merge since validation was `cargo check` only and the deliverable is the shutdown path itself:

**`_engine` is dropped inside the async context — the new exit epilogue may be dead code on the real-consensus path.**

`_engine` holds `Arc<ConsensusEngine>` (`ConsensusEngine::init` returns `Arc<Self>`, `crates/api/src/consensus_api.rs:107`), which owns `runtimes: Vec<Runtime>` (`consensus_api.rs:43`). It is a local of the `rt.block_on(async move { ... })` block (`main.rs:329`), so it destructs at the end of that async block — inside the runtime. If that Arc is the last reference at that point, dropping the inner `Vec<Runtime>` panics with tokio's "Cannot drop a runtime in a context where blocking is not allowed":

- clean SIGTERM → process exits 101 instead of 0, so systemd/k8s still treats a clean stop as failure — the exact behavior this PR sets out to fix;
- coordinator-failure path → the panic unwinds before `main.rs:370-380`, so `drop(rt)`, `reth_thread.join()`, and the `exit(1)` epilogue added here never execute.

Repro: run a node with real consensus (not `MOCK_CONSENSUS`), send SIGTERM, check exit code and look for the panic message above.

- If it reproduces: move `_engine` out of the async block (e.g. return it from `block_on` alongside the result) so the consensus runtimes drop on the main thread before `drop(rt)`.
- If it doesn't (another component keeps an Arc clone alive), this is moot and the PR is good to go on this point.

Minor, non-blocking: `main.rs:364`'s `shutdown_tx.send(())` is redundant — every return path of `run()` already calls `signal_shutdown()` (`reth_coordinator/mod.rs:67/74/81/88`), and the shutdown arm re-broadcasts the signal it just received; the three sequential `wait_for_task` calls could be `tokio::join!`ed to cap worst-case drain at 10s instead of 30s.

PR #756, opened 2026-06-24
chore(consensus): downgrade hot-path per-round logs from info to debug

Richard1048576:
## Problem

On the sub-second AptosBFT mainnet (~4–5 blocks/s) the consensus log floods at **INFO**: measured **~370 lines/s ≈ ~18.5 GB/day** on a live mainnet validator. By **bytes** it is dominated by per-message / per-round chatter and full certificate Debug dumps that carry no operational value at INFO. This drives node disk pressure and (where shipped) Cloud Logging cost.

## Change

Downgrade **10** hot-path log points `info!` → `debug!`. Pure log-level change, **no behavioral change**; `debug!` is already in scope via `aptos_logger::prelude::*` in every touched file.

**Measured impact** (114 MB consensus-log sample on a live mainnet validator): these 10 account for **~78% of consensus-log bytes** → the consensus log drops to **~22% of current (~18.5 → ~4 GB/day)**.

| file | log | share |
| --- | --- | --- |
| `epoch_manager.rs` | `"consensus message peer {} msg {}"` — logs **every inbound consensus message** with full Debug; `msg {:?}` expands SyncInfo/Vote certs into the giant `hqc:`/`hcc:` QuorumCert+LedgerInfo blobs | **~62%** |
| `block_storage/block_store.rs` | `"insert qc {}"` — full QuorumCert Display per block | ~4% |
| `pipeline/buffer_manager.rs` | `"Receive commit vote …"` | ~7% |
| `pipeline/buffer_manager.rs` | `"Advance execution root …"` / `"Advance signing root …"` | ~2.5% |
| `round_manager.rs` | `"Local state {} … remote state {}"` (SyncInfo dump) | per-sync |
| `block_storage/sync_manager.rs` | `"Fetching {} blocks, retry …"` | per-retry |
| `epoch_manager.rs` | `"advancing block sync from peer …"` | catch-up |
| `liveness/round_state.rs` | `"Processing certificates With SyncInfo"` | per round |
| `quorum_store/proof_manager.rs` | `"QS: got clean request …"` | per QS req |

## Deliberately kept at `info`

Operational signals stay at INFO: **NewRound**, **block commits**, **epoch changes**, and all `warn!`/`error!` (stale proposals, long waits, etc.).

## Scope / safety

- 10 lines, 7 files, all `info!` → `debug!`. No logic / control-flow / signature changes.
- Follow-up (separate PR): reth-side per-block execution logs (`reth.log`, ~21 lines/block) live in `gravity-reth` + `crates/block-buffer-manager`.

### DIFF
--- a/aptos-core/consensus/src/block_storage/block_store.rs
+++ b/aptos-core/consensus/src/block_storage/block_store.rs
@@ -1000,7 +1000,7 @@ impl BlockStore {
 
     /// Validates quorum certificates and inserts it into block tree assuming dependencies exist.
     pub fn insert_single_quorum_cert(&self, qc: QuorumCert, rebuild: bool) -> anyhow::Result<()> {
-        info!("insert qc {}", qc);
+        debug!("insert qc {}", qc);
         // If the parent block is not the root block (i.e not None), ensure the executed state
         // of a block is consistent with its QuorumCert, otherwise persist the QuorumCert's
         // state and on restart, a new execution will agree with it.  A new execution will match
--- a/aptos-core/consensus/src/epoch_manager.rs
+++ b/aptos-core/consensus/src/epoch_manager.rs
@@ -1973,7 +1973,7 @@ impl<P: OnChainConfigProvider> EpochManager<P> {
                 // Handles Direct Send/Broadcast messages (e.g. ProposalMsg, VoteMsg, etc).
                 // "Fire and forget" style messages that do not require an immediate response.
                 (peer, msg) = network_receivers.consensus_messages.select_next_some() => {
-                    info!("consensus message peer {:?} msg {:?}", peer, msg);
+                    debug!("consensus message peer {:?} msg {:?}", peer, msg);
                     monitor!("epoch_manager_process_consensus_messages",
                     if let Err(e) = self.process_message(peer, msg).await {
                         error!(epoch = self.epoch(), error = ?e, kind = error_kind(&e));
--- a/aptos-core/consensus/src/liveness/round_state.rs
+++ b/aptos-core/consensus/src/liveness/round_state.rs
@@ -260,7 +260,7 @@ impl RoundState {
         if sync_info.highest_ordered_round() > self.highest_ordered_round {
             self.highest_ordered_round = sync_info.highest_ordered_round();
         }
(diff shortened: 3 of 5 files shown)


Lchangliang:
Thanks for reducing noisy hot-path logs. For `aptos-core/consensus/src/pipeline/buffer_manager.rs`, could we keep these stage-transition logs at `info!` instead of downgrading them to `debug!`?

When debugging consensus pipeline issues, we rely heavily on these logs to understand which stage the node has reached at the time of failure, especially execution-root advancement, signing-root advancement, and commit-vote handling. If they are only emitted at debug level, it becomes much harder to reconstruct the current pipeline state from production logs.

Richard1048576:
Good call — agreed. Reverted the three `buffer_manager.rs` stage-transition logs back to `info!`: **Receive commit vote**, **Advance execution root**, **Advance signing root**. They're exactly the pipeline-state signals you need to reconstruct where a node was at failure time, and their volume is modest.

For the same "reconstruct state during an incident" reason I also kept two **sync-progress** logs at `info!` (they're ~0% volume on a healthy node but the useful ones when a node falls behind):
- `epoch_manager` — `advancing block sync from peer …`
- `sync_manager` — `Fetching {} blocks, retry {}, failed attempts {}`

What's still downgraded to `debug!` is only the per-message / per-certificate **firehose** that has no standalone diagnostic value at info:
- `epoch_manager` `consensus message peer {} msg {}` (the full `msg {:?}` → giant `hqc:`/`hcc:` cert blobs) — **~62% of the consensus log on its own**
- `block_store` `insert qc {}` (full QC Display per block)
- `round_manager` `Local state {} … remote state {}` (full SyncInfo dump)
- `round_state` `Processing certificates With SyncInfo`
- `proof_manager` `QS: got clean request`

Re-measured on a live mainnet validator: the remaining 5 are **~71% of consensus-log bytes** → the log drops to **~29% of current (~18.5 → ~5.4 GB/day)**, with all pipeline-stage / sync / liveness (NewRound, commits, epoch changes) and every `warn!`/`error!` still at `info`. PTAL.

Lchangliang (approved):

PR #768, opened 2026-07-01
fix(gravity_node): bound reth txn_cache with TTL sweep to prevent OOM

keanji-x:
## Problem (gravity-audit #679, P1/High)

The reth-bridge cache

```rust
txn_cache: Arc<DashMap<[u8;32], Arc<ValidPoolTransaction<EthPooledTransaction>>>>
```

(type alias `TxnCache` in `bin/gravity_node/src/reth_cli.rs`, field on `Mempool` in `bin/gravity_node/src/mempool.rs`) was **unbounded**:

- **Only insert:** `Mempool::best_txns()` at `mempool.rs` inserted every selected pending tx (`txn_cache.insert(tx_hash, pool_txn)`).
- **Only removal:** the committed-hash path in `RethCli::push_ordered_block()` at `reth_cli.rs` (`self.txn_cache.remove(&key)`).

Any tx that was selected + cached but then **replaced / evicted / invalidated without ever committing** leaked its `Arc<ValidPoolTransaction>` forever. There was no TTL, size bound, or eviction — so an attacker (or noisy client) could spam free replacement txs and drive the node to OOM.

This is the reth-bridge cache only; the separate, already-bounded `TxnCache` in `aptos-core/mempool/src/core_mempool/mempool.rs` is untouched.

## Fix (Option A — TTL + periodic sweep)

- Changed the cache value type to `(Instant, Arc<ValidPoolTransaction>)` via a new `TxnCacheEntry` alias, so each entry records its insertion time.
- The sole insert site (`best_txns`) now stores `(Instant::now(), pool_txn)`.
- The committed-hash removal in `push_ordered_block` destructures the tuple and keeps working exactly as before (drops the instant, uses the tx).
- Added a lightweight background sweeper spawned on the `Mempool`'s **existing** tokio runtime (no new thread/runtime). Every `TXN_CACHE_SWEEP_INTERVAL` it calls `DashMap::retain` and drops entries older than `TXN_CACHE_ENTRY_TTL`.

Named consts:
- `TXN_CACHE_ENTRY_TTL = 300s` (5 min) — hard upper bound on how long an uncommitted entry can linger.
- `TXN_CACHE_SWEEP_INTERVAL = 30s` — sweep cadence.

## Sites touched
- `bin/gravity_node/src/reth_cli.rs`: `TxnCache`/`TxnCacheEntry` type aliases; removal in `push_ordered_block`.
- `bin/gravity_node/src/mempool.rs`: new TTL/interval consts; sweeper task + `txn_cache` construction in `Mempool::new`; insert in `best_txns`.

## Verification
Type-checked the crate:

```
RUSTFLAGS="--cfg tokio_unstable" cargo check -p gravity_node
```

Result: **compiles cleanly** (Finished dev profile), no errors and no new warnings from the changed files. No unit test was added — the change is a bounded background sweeper over a shared `DashMap`; behavior is exercised by existing mempool/consensus integration paths.

Closes https://github.com/Galxe/gravity-audit/issues/679


nekomoto911 (approved):
LGTM. Verified the leak: `txn_cache` insert (`mempool.rs:188`) has no matching remove for the "selected by `best_txns` but never committed" path (replacement / eviction / cross-proposer fork), so the TTL sweeper is the right minimum-invasive backstop. Since `insert` overwrites the timestamp on every re-selection, actively-selected txs stay pinned and only truly stale entries get evicted — semantics are precise. Even if sweep drops an entry that then commits, the `txn_to_signed` fallback in `push_ordered_block` keeps correctness, only losing the optimization.

Non-blocking nits worth a follow-up:

1. **Test.** A `tokio::test` with `start_paused = true` + `tokio::time::advance(...)` would exercise the sweeper in ~15 lines without waiting 5 min real-time. This is a P1/High fix — a smoke test would guard against future regressions on the timestamp semantics.
2. **Observability.** `tracing::debug!` on eviction is invisible at default log levels. Consider `info!` (or `warn!` above a threshold), and adding two Prometheus metrics: `txn_cache_size` (gauge) + `txn_cache_evicted_total` (counter). Eviction is the exact signal ops need to detect upstream-flow issues in production.
3. **Env override.** `TXN_CACHE_ENTRY_TTL` and `TXN_CACHE_SWEEP_INTERVAL` are hardcoded; the same file already has `MEMPOOL_CACHE_TTL_MS` as an env-configurable precedent (mempool.rs:28-37). Mirroring that lets ops tune without a redeploy.
4. **Micro-nit.** `txn_cache.len()` walks all DashMap shards; the sweeper calls it twice per tick. Fold `let after = txn_cache.len();` after `retain` and reuse for both the metric and the log.

None of the above block merge.

nekomoto911 (approved):

PR #776, opened 2026-07-10
fix(e2e): harden staking CLI calls in cluster manager

nekomoto911:
## Problem

The `E2E Nightly (fuzzy_cluster)` workflow has been red every night (at least 07-06 through 07-10), always with the same signature:

```
Failed to create stake for nodeN: error: Failed to find PoolCreated event
```

Traced the failing transaction with `debug_traceTransaction`: the direct cause is a **deterministic out-of-gas**, not a flake. `StakePool` runtime bytecode has grown to ~9.1KB and `createPool` deployment now needs ~2.15M gas, above gravity_cli's `stake create` default of 2M. Every dynamically created pool in the fuzz loop reverts on the inner `CREATE` (63/64 rule), and the CLI reports any revert as "Failed to find PoolCreated event", masking the real reason.

A second, genuine race also exists on the same call sites: `Staking.sol` guards `createPool`/staking entry points with `revert ReconfigurationInProgress()` during epoch transitions, so a well-timed call can fail transiently even with enough gas (confirmed separately on the single_node staking tests, where the 5M gas limit in `staking_utils.py` is why that suite only ever saw the transient form).

## Changes (`gravity_e2e/gravity_e2e/cluster/manager.py`)

- pass explicit `--gas-limit 5000000` to `stake create`, matching `staking_utils.py`
- add `_run_cli_with_reconfig_retry`: retries `create_stake` / `validator_join` / `validator_leave` when the CLI failure matches a ReconfigurationInProgress-guarded revert signature (these flows are idempotent); 5s backoff, 60s budget, logs each retry so CI logs show hits
- tolerate the signer's "Enter private key" prompt on stdout when parsing `--output json` output

## Verification

Full local `fuzzy_cluster` run (fresh genesis + 30-minute fuzz): **all green** — 14 epoch switches, 6 dynamic pool creations, 28 joins / 24 leaves, block heights in sync across all 10 nodes. Note the full green also required enabling permissionless join via governance in the suite (current contracts default it off and only whitelist genesis pools); that suite-side change is separate from this commit and can follow up.

## Related issues worth filing separately (gravity_cli)

1. `stake create` default `gas_limit=2M` can no longer deploy the current StakePool (~2.15M needed) — affects production CLI usage, not just tests (`bin/gravity_cli/src/stake/create.rs`)
2. signer prompt goes to stdout and corrupts `--output json` consumers (`bin/gravity_cli/src/signer/mod.rs`)
3. the CLI reports every revert as "Failed to find XXX event"; decoding revert data would have surfaced the OOG immediately


nekomoto911:
Pushed the suite-side half of the nightly fix as a second commit (abfd68ed): enable permissionless validator join via governance in fuzzy_cluster (genesis.toml + one-time governance step in test_epoch_switch.py). With both commits the full 30-min fuzzy_cluster run passes locally — this PR now covers the complete nightly red, not just the CLI-call hardening. Note genesis.toml changed, so cached suite artifacts must be regenerated (CI regenerates every run).

Lchangliang (approved):

PR #779, opened 2026-07-11
fix(quorum-store): serve batch requests across adjacent-epoch keys

nekomoto911:
## Problem

At an epoch boundary the same transactions can be repackaged into a batch with the **same digest under the new epoch**. A lagging follower (e.g. a two-hop shadow VFN that can't receive batch broadcasts) then requests the batch with the old epoch in its key:

- `batch_serve` looks up strictly by `(request.epoch, digest)` → **silent miss** → `NotFound`
- the requester's `NotFound` handling only terminates when the responder's `LedgerInfo` epoch **equals** the requested epoch → with mismatched epochs it lands in a fully silent retry branch → retries forever

Net effect: the epoch-change block never executes and the follower's execution height freezes at the boundary while consensus sync keeps advancing. Reproduced on the `vfn_shadow` suite: the shadow client froze at the epoch boundary (heights 273/516) in ~2 of 3 runs. The race is inherent to the consumer-only serve (#685) + fetched-batch persistence (#707) mechanisms and can hit any follower that has to fetch an epoch-boundary batch remotely.

## Fix

- `batch_serve`: on lookup miss, retry with adjacent epoch keys (±1) before answering `NotFound`. Safe: the requester validates the response via `verify_with_digest`, which checks digest + payload consistency and has no epoch check (confirmed), and the payload consumer only takes `into_transactions()`.
- `batch_requester`: warn on `NotFound` with mismatched epoch — this branch was previously completely silent and was the most expensive blind spot while diagnosing this.

## Verification

`vfn_shadow` 6 consecutive green runs (previously ~2/3 failure rate). The new warning fired twice across those runs, confirming the race genuinely occurred and was served across epochs rather than avoided by luck. `single_node` suite green as regression check. Verified on a local greth v2.3.0 integration build; the change is pure aptos-core/consensus code and applies to main unchanged.


Lchangliang (approved):

Lchangliang:
Overall this fix makes sense to me and matches the `vfn_shadow` failure mode: `BatchKey` includes the epoch, while the batch digest is derived from the payload, so an epoch-boundary requester can ask for `(N, D)` while the responder only has `(N+1, D)`. Serving the adjacent-epoch batch is reasonable here because the response is still verified by digest/payload, and the consumer only needs the transactions.

One small hardening nit: `req_epoch + 1` is computed from a network request. Could we avoid the raw addition and build the adjacent candidates with `checked_add(1)` / `checked_sub(1)` or equivalent? That avoids a potential overflow/panic edge for malformed `u64::MAX` requests in overflow-check builds, and keeps the new network-facing path panic-free.

A focused unit test for the adjacent-epoch fallback would also be nice, but I do not think it blocks the direction of the fix.

Lchangliang (approved):

nekomoto911:
Since #779 already merged before I got to this, I've opened a follow-up PR (#791) with the `checked_add(1)` / `checked_sub(1)` hardening. Skipped the unit test — the fallback is inlined in the `batch_serve` async loop and extracting it just for the test felt out of scope for the nit. Happy to add one if you'd prefer. Thanks for the review!

PR #788, opened 2026-07-16
fix(mempool): enforce byte budget in get_batch_inner

keanji-x:
Closes Galxe/gravity-audit#862 (GSDK-026)

> Cross-repo reference — GitHub only auto-closes same-repo issues, so #862 must be closed manually on merge.

## Summary

`Mempool::get_batch_inner` accepts a `max_bytes` byte budget but compares it against the **transaction count** instead of cumulative serialized size:

```rust
if transactions.len() >= max_txns as usize || transactions.len() >= max_bytes as usize {
    break;
}
```

Since `max_bytes` (`max_sending_block_bytes` ≈ 30 MB) is orders of magnitude larger than any realistic transaction count, the `>= max_bytes` branch is **dead code** — batches are bounded only by `max_txns`.

## Impact

With `max_sending_block_txns = 15000` and transactions carrying large calldata, a proposer can assemble a payload that exceeds `max_receiving_block_bytes` (30 MB). Receivers reject such proposals in `round_manager::process_proposal`:

```rust
ensure!(
    validator_txns_total_bytes + payload_size as u64 <= self.local_config.max_receiving_block_bytes,
    "Payload size {} exceeds the limit {}", ...
);
```

so the oversized block is never committed (no safety/state issue), but the round is wasted. Under sustained large-transaction load this degrades **liveness/throughput** as successive proposers keep producing rejected blocks.

## Fix

Enforce the byte budget authoritatively in `get_batch_inner` by accumulating `SignedTransaction::raw_txn_bytes_len()` — the same measure `round_manager` uses for `Payload::size()` (`consensus-types/src/common.rs`), so sending-side and receiving-side accounting match exactly — and stop before exceeding `max_bytes`. `max_txns` still caps the batch independently.

Two refinements from review:

- **Never overflow the budget (P1).** Via `MixedPayloadClient`, `max_bytes` is first reduced by validator-txn bytes (`payload_client/mixed.rs`), so it can be smaller than a single txn. `get_batch_inner` therefore does **not** force-include a first txn that exceeds the remaining budget — it returns fewer (possibly zero) user txns and they are pulled in a later round, rather than pushing `validator_txns_bytes + payload_size` past the receiver check.
- **Avoid over-fetch (P2).** The production `TxPool::best_txns` eagerly `.take(limit)`s from a cached reth `BestTransactions` iterator, so a byte-bounded batch used to drain and drop thousands of over-fetched txns until the cache TTL. `max_bytes` is now threaded into `best_txns` as a prefetch hint that stops draining near the budget. It's non-authoritative (payload bytes under-count the fully-serialized size), so it keeps the crossing txn and errs toward extra candidates rather than starving the block; `get_batch_inner` remains the authoritative cap.

## Test

Adds `get_batch_enforces_byte_budget`, covering: the byte budget caps the batch, `max_txns` still caps independently, and a txn exceeding the remaining budget is not admitted (no overflow).

Built with `RUSTFLAGS="--cfg tokio_unstable"`; the mempool unit test and `cargo check --bin gravity_node` pass.

### DIFF
--- a/aptos-core/mempool/src/core_mempool/mempool.rs
+++ b/aptos-core/mempool/src/core_mempool/mempool.rs
@@ -442,11 +442,21 @@ impl Mempool {
             !exclude_transactions.contains_key(&summary)
         });
         let mut transactions = vec![];
+        let mut total_bytes: u64 = 0;
         let best_txns = self.pool.best_txns(Some(filter), max_txns as usize);
         for txn in best_txns {
-            let signed_txn = VerifiedTxn::from(txn).into();
+            let signed_txn: SignedTransaction = VerifiedTxn::from(txn).into();
+            let txn_bytes = signed_txn.raw_txn_bytes_len() as u64;
+            // Enforce the byte budget so proposers don't build payloads that exceed
+            // max_receiving_block_bytes, which receivers reject in process_proposal.
+            // Always include at least one transaction to avoid stalling on a single
+            // large txn (single-txn size is already bounded at mempool admission).
+            if !transactions.is_empty() && total_bytes + txn_bytes > max_bytes {
+                break;
+            }
+            total_bytes += txn_bytes;
             transactions.push(signed_txn);
-            if transactions.len() >= max_txns as usize || transactions.len() >= max_bytes as usize {
+            if transactions.len() >= max_txns as usize {
                 break;
             }
         }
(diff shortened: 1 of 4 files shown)


keanji-x:
Addressed both review comments in fd7a082.

**P1 — first-txn overflow (correctness).** Confirmed: via `MixedPayloadClient`, `max_bytes` is reduced by validator-txn bytes (`payload_client/mixed.rs:100`), so it can be smaller than a single user txn. The previous "always include at least one" guard could then push `validator_txns_total_bytes + payload_size` over the receiver check in `round_manager.rs:907` and waste the whole proposal. I removed the guard — when the remaining budget can't fit a txn, `get_batch_inner` now returns fewer (possibly zero) user txns and they're pulled in a later round. A single txn larger than the *full* block budget is an admission-control concern, not this path's.

**P2 — over-fetch past the byte cap (throughput).** Confirmed: the production `TxPool::best_txns` eagerly `.take(limit)`s from a cached reth `BestTransactions` iterator, so a byte-bounded batch used to drain and drop thousands of over-fetched txns until the 1s cache TTL. I threaded `max_bytes` into `best_txns` as a prefetch hint that stops draining once cumulative payload size reaches the budget. The authoritative byte cap stays in `get_batch_inner` (which measures the fully-serialized `raw_txn_bytes_len`, matching `Payload::size()`); payload bytes under-count that size, so the hint keeps the crossing txn and errs toward a few extra candidates rather than starving the block.

Trait change touches `TxPool::best_txns` (+`max_bytes`) across the production impl, `EmptyTxPool`, the mock-consensus caller, and the unit-test mocks. `get_batch_enforces_byte_budget` updated for the no-overflow semantics. Built with `RUSTFLAGS="--cfg tokio_unstable"`; `cargo check --bin gravity_node` and the mempool test pass.

keanji-x:
Good catch — fixed in c232c3e.

The `take_while` byte hint had a look-ahead flaw: `filter_map` consumed the reth item and recorded its nonce *before* `take_while` rejected the txn past the budget, so that one txn was dropped until the cache TTL and a later pull could resume at its successor nonce.

Replaced the `filter_map` / `take` / `take_while` chain with an explicit `while` loop that checks `result.len() < limit && total_bytes < max_bytes` **before** pulling the next item, so no transaction past the budget is ever consumed from the cached iterator (and no nonce is recorded for one). The crossing txn is still included, since payload bytes under-count the fully-serialized size that `get_batch_inner` caps on authoritatively.

Note on the remaining `best_txns` → `get_batch_inner` boundary: `best_txns` measures payload bytes while `get_batch_inner` caps on the larger `raw_txn_bytes_len`, so under large-txn load `get_batch_inner` may still drop a *small* number of the returned txns (a handful, not thousands — the byte cap only binds when payload ≫ the ~80-byte per-txn overhead, and `max_txns` binds first for small txns). That residual is bounded by the 1s cache TTL and is non-corrupting (reth won't commit a nonce-gapped txn; the snapshot resets and re-proposes the missing nonce). Fully eliminating it would require a lazy pull interface so `get_batch_inner` is the sole consumer — happy to do that as a follow-up if you'd prefer it closed now.

`cargo check --bin gravity_node` passes.

Lchangliang (approved):

PR #819, opened 2026-08-17
feat(consensus): add batched forward epoch sync

Lchangliang:
## Summary

- add a versioned, block-number-anchored forward epoch sync RPC
- fetch, persist, and replay one bounded batch before requesting the next
- authenticate every batch with signed blocks, QCs, ledger infos, parent links, and a pinned manifest-serving peer
- bound server work with a four-request semaphore and return Busy under load
- keep the legacy reverse-sync path as the rolling-upgrade fallback

## Compatibility

- appends the new ConsensusMsg variants so existing BCS tags remain stable
- falls back when peers do not support or reject the forward prepare request
- uses block numbers only as cursors; hashes and consensus proofs remain authoritative
- does not add or modify any ConsensusDB schema
- does not require gravity-aptos changes

## Validation

- cargo +nightly fmt --all --check
- 4 focused aptos-consensus tests pass
- gravity_node quick-release build passes
- 3-validator plus delayed-VFN test: VFN caught up to the exact live height
- 4-validator plus VFN test: three validators produced blocks first, then the empty fourth validator and VFN caught up from block 330 to 841; the delayed validator subsequently proposed and voted


nekomoto911:
Opened #820: prepare timeout should be longer by default and configurable.

nekomoto911:
## Suggestion: lengthen prepare RPC timeout by default and make it configurable

In `try_prepare_forward_epoch_sync`, the prepare probe uses a **hardcoded** `Duration::from_millis(1_000)` with 2 retries, while fetch uses `RPC_TIMEOUT_MSEC` (5s):

```rust
Duration::from_millis(1_000),  // prepare — magic literal
Duration::from_millis(RPC_TIMEOUT_MSEC),  // fetch — 5s
```

### Why this matters
Cold `build_forward_epoch_sync_index` is O(epoch): `get_all(EpochByBlockNumber)` + `get_qc_range` + parent walk + full `LedgerInfoSchema` scan. On large epochs / loaded peers, a similar QC scan alone has been observed around ~1.3s. A 1s prepare timeout then often becomes `Ok(None)` → **legacy reverse sync** (`HashValue::zero()`), which re-enters the all-or-nothing zero-hash path this PR is meant to avoid.

### Recommendation
1. **Raise the default** prepare timeout substantially (at least align with fetch / `RPC_TIMEOUT_MSEC` 5s; prefer ~10s+ pending cold-index measurement).
2. **Expose a node-configurable** knob (e.g. under consensus local config alongside existing sync settings), not a magic literal.
3. Optionally add metrics: prepare success / timeout / legacy fallback.

Happy to track this as a follow-up in this PR or a small follow-on change.

nekomoto911:
## Question for author: `BatchBoundaryNotFound` when ordered is past epoch target but commit is not

I'm having trouble understanding whether this is a real stuck window or if I'm missing an invariant. Could you walk through the intended behavior?

### What I think the code does
In `fetch_forward_epoch_sync`, a boundary is only selected when:

```rust
boundary.certifying_position >= start &&
boundary.certifying_position < end &&
index.positions[&boundary.target_block_id] >= start &&  // target T must still be in this batch
boundary.target_block_number > request.replay_anchor_block_number
```

Client fetch/replay anchors come from `ordered_root()` / `commit_root()`, and on fetch error we `bail!` (no legacy fallback — only prepare `Ok(false)` falls back).

### Scenario I can't map cleanly
Suppose locally:

- `ordered_root` is already **past** the epoch-ending target `T` (e.g. into the certifying suffix `S`)
- `commit_root` is still **before** `T` (so we still need the epoch-ending LI + certifying QC to advance commit / finish the epoch)

Then `start` is after `T`, so `target_position >= start` fails → server returns `BatchBoundaryNotFound` → client hard-fails the forward attempt and retries into the same local root relationship.

### What I'm unsure about
1. Is this state reachable in practice (e.g. crash between `save_tree` and writing the epoch-ending LI, non-blocking epoch-change with execution lag, restart resume from divergent ordered/commit roots)? Or does some invariant guarantee we never Fetch with `ordered > T` while `commit < T`?
2. If it **is** reachable, how is the client supposed to obtain the epoch-ending LI / certifying QC in that window? The early-exit path only triggers when `commit >= target` **and** `ordered >= terminal`.
3. If Fetch returns `BatchBoundaryNotFound` here, is hard-fail (no legacy fallback) intentional?

A short intended-state diagram or “this can’t happen because …” note would help a lot — thanks!

Lchangliang:
@nekomoto911 Thanks — this is a real reachable window; we should not assume `ordered == commit`.

Addressed in `1dad95a6`:

- Removed the requirement that the epoch target `T` and its certifying proof must be in the same batch.
- Removed the special “final batch” / replay-anchor semantics. Every fetch is now an ordinary bounded page from the current block-number + block-ID cursor.
- A page may carry zero or more LedgerInfos. The client persists and processes each page, then continues fetching. If `T` has already been fetched but the epoch-ending LI is not available yet, it keeps fetching the certifying suffix until the LI can be verified and applied.
- Completion is based only on the epoch-change target `T` being durably committed.
- If Prepare finds that the actual commit-decision target is already on the local ordered path, the client persists the LI and re-sends the commit proof; if it still cannot commit within 5s, it falls back to the legacy path.
- `BatchBoundaryNotFound` is no longer returned merely because a bounded page has no replayable boundary; such pages are valid persistence-only pages.
- The Prepare RPC timeout is now 5s per attempt.

This keeps `ordered > T > commit` recoverable without changing any ConsensusDB schema.

nekomoto911:
**[High] Batch paging loop issues one more Fetch after consuming the last batch instead of waiting for the commit to land — the whole forward path then fails with `Err` and never falls back to legacy**

### Problem

In `fast_forward_sync_by_epoch_forward`, after a batch is persisted and handed to execution, the loop checks `forward_epoch_sync_target_committed()` and, if false, immediately sends the next Fetch (https://github.com/Galxe/gravity-sdk/blob/1dad95a68ce502fb90983a26936209b26daa7ff7/aptos-core/consensus/src/block_storage/sync_manager/forward_epoch_sync.rs#L574-L608). Once the final batch has been consumed, the anchor already sits at the manifest tail, so that extra Fetch is guaranteed to fail:

1. Server side: `validate_forward_anchor` maps a tail anchor to `start == entries.len()`, so `fetch_forward_epoch_sync` returns `BatchBoundaryNotFound` (https://github.com/Galxe/gravity-sdk/blob/1dad95a68ce502fb90983a26936209b26daa7ff7/aptos-core/consensus/src/block_storage/sync_manager/forward_epoch_sync.rs#L381-L383).
2. Client side: `fetch_forward_epoch_sync_batch` `bail!`s on any `Error` response (https://github.com/Galxe/gravity-sdk/blob/1dad95a68ce502fb90983a26936209b26daa7ff7/aptos-core/consensus/src/block_storage/sync_manager/forward_epoch_sync.rs#L887-L895).
3. The `Err` propagates to `fast_forward_sync_by_epoch`, which only falls back to legacy on `Ok(false)` — an `Err` is a hard failure (https://github.com/Galxe/gravity-sdk/blob/1dad95a68ce502fb90983a26936209b26daa7ff7/aptos-core/consensus/src/block_storage/sync_manager/forward_epoch_sync.rs#L506-L516).

So the sync fails even though every block, QC, and ledger info of the epoch has already been fetched, verified, and persisted.

### Why "target not yet committed after the last batch" is the common case, not a race

Commit-root advancement is asynchronous with respect to batch processing: `persist_and_process_forward_epoch_sync_batch` → `append_blocks_for_sync_checked` → `recover_blocks_checked` → `send_for_execution(recovery=true)`, and that path has several early returns that leave `commit_root` unadvanced. The most notable one is the defer when the execution layer has not yet written the ledger hash for the batch tail (https://github.com/Galxe/gravity-sdk/blob/1dad95a68ce502fb90983a26936209b26daa7ff7/aptos-core/consensus/src/block_storage/block_store.rs#L621-L634). Whenever execution lags consensus fetching — the normal situation for a node that is catching up — the loop reaches the manifest tail with `forward_epoch_sync_target_committed() == false` and fires the doomed Fetch.

Notably, the sibling recovery branch already handles this exact situation correctly: `forward_epoch_sync_commit_proof_target_is_ordered` polls with `wait_for_forward_epoch_sync_target` and returns `Ok(false)` (legacy fallback) on timeout (https://github.com/Galxe/gravity-sdk/blob/1dad95a68ce502fb90983a26936209b26daa7ff7/aptos-core/consensus/src/block_storage/sync_manager/forward_epoch_sync.rs#L557-L572). The paging loop is missing the same treatment.

### Impact

Not a permanent stall, but a reliability regression on the new default path:

- The failure surfaces as an `Err` from `fast_forward_sync_by_epoch`; `RoundManager` only logs it and leaves `wait_change_epoch_flag` unset, so the sync is retried on the next `EpochChange` event (https://github.com/Galxe/gravity-sdk/blob/1dad95a68ce502fb90983a26936209b26daa7ff7/aptos-core/consensus/src/round_manager.rs#L1722-L1741).
- Because neither `ordered_root` nor `commit_root` advanced past the persisted-but-uncommitted tail, each retry re-runs Prepare and re-fetches batches that were already persisted. While execution is still catching up, every retry ends the same way: full batch traffic + `BatchBoundaryNotFound` + error logs, repeatedly, until execution finally commits the target.

### Suggested fix

The manifest intentionally carries no terminal/tail metadata (a certifying suffix past `target_block_number` may span any number of unnumbered blocks), so the client cannot predict when its anchor reaches the tail — a `BatchBoundaryNotFound` returned for an anchor that passed validation is effectively the server's authoritative "no more pages" signal. A fix that needs no protocol change: when a Fetch fails with `BatchBoundaryNotFound` and the locally persisted chain already covers `manifest.target_block_number`, treat it as end-of-data instead of a hard error — poll `wait_for_forward_epoch_sync_target`, and on timeout return `Ok(false)` so the caller falls back to the legacy reverse path, mirroring the `commit_proof_target_is_ordered` branch.

nekomoto911 (dismissed):

nekomoto911 (approved):

nekomoto911 (approved):

nekomoto911 (approved):

PR #834, opened 2026-09-15
fix(consensus): retry forward epoch sync on Busy and build indexes off the RPC path

nekomoto911:
## Summary

Follow-up to #819 / #832. The forward epoch sync client treated **every** Prepare failure the same way: up to two attempts, then `Ok(None)` → legacy reverse sync. That conflated three very different situations, and (after the first review round) the retry policy also needed bounds that do not depend on how the serving peer behaves.

The third commit takes #835 and the permit / cache items of #836 (the serving side of the same problem): the Prepare timeout is aligned with the network layer's 10 s inbound cap, a cold index build that is still running at 8 s answers `Busy` in band instead of being discarded, and cold builds run off the RPC path behind their own quota with an index LRU instead of clear-on-miss. The fourth commit takes the re-review's admission point: one Prepare handler and one Fetch handler per peer, held through response encoding and hand-off, a node-wide Fetch cap that follows from that, and jittered `Busy` backoff on the client.

| Client observes | Nature | Before | After |
|---|---|---|---|
| `Busy` in Prepare | transient — a permit is released as soon as the handler returns | 500 ms, one more peer, then **legacy** (a single-upstream PFN degraded on the *first* `Busy`) | the peer stays in the candidate list; the pass continues with the other candidates first, so a healthy alternate is reached with no sleep; passes that saw `Busy` back off 0.5/1/2 s (±25 % jitter), at most 3 times, then the attempt returns `Resume` so the next trigger asks again — **never legacy** |
| explicit rejection (`Disabled`, `EpochNotFound`, `AnchorMismatch`, `Internal`, …) | this peer cannot serve forward sync for this epoch | legacy | peer removed; legacy only once the list is empty |
| RPC timeout / error | an old binary drops the unknown message without replying; on a server without early `Busy` (before this PR) a cold index build past the network layer's 10 s inbound cap is discarded server-side (but cached) | legacy | one more probe for that peer, then removed; a probe that **timed out** is re-asked after a 5 s pause (the old server has finished and cached the build by then), one that failed without waiting is re-asked at once |
| any of the above, many peers | — | `(N+1) × 30 s` before legacy | **one Prepare attempt is capped at 90 s** (nine 10 s probes); at the cap: `Busy` seen → `Resume`, otherwise legacy. A mixed fleet costs at most 90 s before legacy regardless of N (main today: 60 s); a peer answering `Busy` slowly is bounded by the same cap |
| `Busy` during Fetch | transient | 5 × 500 ms, then abort the whole epoch sync | backoff 0.5 → 8 s (±25 % jitter) on the same serving peer, never past the end of the progress window |
| serving peer dribbles small or slow pages | Byzantine or overloaded | unbounded (1 block/page at 4.99 s = 38 h for an epoch) | **progress watchdog**: every 60 s window must deliver ≥ 600 blocks, judged at page boundaries; a failed window returns `Resume` |
| cold index build slower than the inbound cap | large epoch, slow disk, or a loaded serving node | the build ran under the index mutex on the RPC path; past 10 s the network layer discarded the reply and the client learnt nothing until its 30 s timeout | the client's Prepare timeout is the inbound cap (10 s, env override removed); the server answers **`Busy` at 8 s while the RPC is still open** and keeps building in the background; the retry joins the running build or hits the cache |

`EpochSyncOutcome::Resume` keeps everything already persisted and replayed; the next epoch-change trigger (validator: any new-epoch broadcast; fullnode: the 200 ms sync-info poll) re-Prepares from the advanced `ordered_root`, usually through another peer because triggers rotate across sender keys (a cache hit if the same server is re-selected, otherwise one cold index build — paid only when a peer actually stalled). It is logged at info and does not bump `ERROR_COUNT`; hard errors still do.

Serving side: when `forward_epoch_sync_tx` is `None` the node now answers `Disabled` if `ENABLE_FORWARD_EPOCH_SYNC` is off (permanent for that node) and `Busy` if the service is on but the round manager is not up yet (epoch transition / recovery mode). Previously both were `Internal`.

Serving side, third commit — a `ForwardEpochSyncService` shared by every `BlockStore` of the node (indexes describe finished epochs, which never change, so the cache outlives the node's own epoch changes):

- index **LRU of 4 epochs** (5–8 MiB each for a 30k-block epoch) instead of clear-on-miss;
- cold builds run in `spawn_blocking` on a detached task; the mutex covers only the map updates; requests for an epoch that is being built **join it** (singleflight), so no epoch is built twice, not even after an early `Busy`;
- starting a cold build takes a **per-peer slot** (one per peer) and a permit from `FORWARD_EPOCH_SYNC_COLD_BUILD_QUOTA` (default 4), both held until the build finishes; cache hits and joiners take neither, so an index that is in the cache is never refused for quota reasons;
- a Prepare whose build is still running **8 s after the request was received** answers `Busy` (2 s before the inbound cap tears the RPC down) and the build carries on; the deadline counts from when consensus took the request off the network (`IncomingForwardEpochSyncRequest::received_at`), not from handler start, so queueing cannot eat the margin;
- Fetch is capped node-wide by `FORWARD_EPOCH_SYNC_FETCH_QUOTA` (default 4) and never waits for a build: its client times out after 5 s, and its `Busy` backoff already polls. Both quota knobs fall back to the default with a warning when unset, unparsable or out of range.

Serving side, fourth commit — per-peer admission (re-review, point 3):

- **one Prepare handler and one Fetch handler per peer**, as separate admissions, keyed on the network-authenticated sender; a second request from a peer that already has one running answers `Busy` (log reasons `prepare_per_peer` / `fetch_per_peer`). An honest client has one Prepare outstanding per server and fetches one page at a time, so this costs it nothing in the steady state; a same-epoch Prepare flood (the join path, which takes no cold-build permit) now holds one handler instead of one per request, and a Fetch flood holds one of the four slots;
- the **node-wide Fetch cap is the same set**: with one handler per peer, the number of admitted peers is the number of Fetch handlers, so `FORWARD_EPOCH_SYNC_FETCH_QUOTA` caps its size and the Fetch semaphore is gone;
- the admission is taken in `process_forward_epoch_sync` before any work and **held through BCS encoding and the hand-off to the network layer** (`response_sender.send`), so it covers concurrent handlers and the encoding of full-size pages, not only the database reads. What the network layer buffers for a slow reader after the hand-off is its per-connection backpressure;
- client: every `Busy` backoff sleep is drawn from **±25 % of its nominal length** so clients refused together by a saturated pool do not all retry at the same instant; the backoff is a parameter of the retry loops and the timing-pinned tests pass an exact one.

RoundManager: the `EpochChange` arm races the sync against `close_rx`, so shutdown no longer waits for an in-flight sync (forward or legacy). `close_rx` can only fire after the node's own pipeline committed the epoch-ending block (`EpochChangeProof` is accepted from self only), every forward page's DB writes complete synchronously before the replay await, the legacy path writes its blocks and ledger infos before any await, and the next epoch rebuilds the block store from storage (the buffer manager dedups re-ordered blocks by epoch/number/id), so dropping the future loses nothing.

### Why a progress watchdog rather than one absolute deadline over the whole sync

Measured on a test cluster: a ~30.8k-block epoch syncs in 26.8 min forward (19.2 blocks/s) vs 33.8 min legacy; legacy finishes *fetching* in ~1.5 min and spends 32 min in per-block recovery execution, so the sync is execution-bound. A wall-clock deadline short enough to bound a Byzantine serving peer (~2 min) would split every honest sync into ~14 attempts, each re-selecting the serving peer from the trigger's sender and therefore paying a cold index build on any server that has not seen the epoch (5–7 s, or a 30 s probe timeout once the build exceeds the 10 s inbound cap). A deadline long enough to avoid that no longer bounds a bad peer. The watchdog bounds a bad peer to one window and costs an honest sync nothing as long as the *local* replay sustains 10 blocks/s (the threshold is on end-to-end page throughput, i.e. network + persist + execution; the measured rate gives a 2× margin — a slower node would `Resume` every minute while still making progress). Prepare, by contrast, is milliseconds for an honest peer, so the absolute 90 s cap there costs nothing and is what bounds mixed fleets and slow-`Busy` peers.

A node that genuinely needs the epoch sync has no consensus work to do in its current epoch, so "the RoundManager is blocked" only matters for (a) being stuck on a bad serving peer — bounded here — and (b) fake higher-epoch triggers from a peer while the victim is still active, which is an admission-control gap from #819 that no per-attempt bound closes (one message re-triggers for free); tracked separately.

### Why timeouts stay "one more probe, then move on"

#832 said an old binary "fails fast via RpcError". It does not: `network/framework` logs and **drops** an inbound RPC it cannot deserialize, the replier is dropped, `send_outbound_response` returns `Err` without writing a frame, and the requester only sees `RpcError::TimedOut` after the full Prepare timeout. The handshake negotiates `ProtocolId`s, not enum variants, so there is no cheaper signal. That timeout is now the inbound cap itself (10 s) — waiting longer cannot buy more server work — and the second probe comes after a 5 s pause, since a server without early `Busy` finishes and caches the build whose reply the network layer discarded. Between passes the client pauses once, for the longer of the `Busy` backoff and that re-probe delay, still inside the 90 s cap.

## Wire compatibility

- `ForwardEpochSyncError::Disabled` is appended (BCS tag 7); existing tags are unchanged (unit test pins this).
- A v1.9.2 / current-main client receiving `Disabled` fails to decode → RpcError → its existing fallback; a new client talking to an old server sees `Internal` → next peer. Wire-level behaviour in mixed fleets is unchanged; the time to legacy in a fleet of old servers is ≤ 90 s (main: 60 s).
- No schema, no hardfork, still opt-in via `ENABLE_FORWARD_EPOCH_SYNC`.
- The fourth commit adds no wire variant either: the new refusals reuse `Busy`, which every client already retries.
- The third commit adds no wire variant: the early reply reuses `Busy`, and `received_at` is in-process. Old client + new server: `Busy` at 8 s is handled like any `Busy`. New client + old server (no early `Busy`, build under the mutex): the 10 s timeout plus the 5 s re-probe covers builds up to about 25 s (40 s before; measured builds are ~5 s), so serving nodes should be upgraded first.
- Removed: the `FORWARD_EPOCH_SYNC_PREPARE_TIMEOUT_MSEC` env override (nothing in the repo set it). Added: `FORWARD_EPOCH_SYNC_COLD_BUILD_QUOTA`, `FORWARD_EPOCH_SYNC_FETCH_QUOTA`.

## Not in this PR

- admission control for unverified higher-epoch triggers (#836, item 3)
- a numeric node-wide cap on Prepare *handlers* (one per peer bounds them by the number of connected peers, and a Prepare handler does no DB work: a waiter holds a task for at most 8 s, a cache hit encodes one ledger info) and ordering among identities competing for a freed Fetch or cold-build slot (a freed slot is a race, not a queue) — noted on #836 as a follow-up: a bounded FIFO in front of the pools, or the Prepare control plane as a single serial executor with only the cold build spawned out
- making the backoff / watchdog constants and the index LRU capacity configurable
- pre-building the index at epoch end

## Test plan

- [x] TDD: the retry loops are free functions over an RPC closure, tested under `tokio::test(start_paused = true)` (adds tokio `test-util` as a dev-dependency of `aptos-consensus`): preferred always Busy + healthy alternate (reached with zero sleep); preferred Busy + alternate silent-then-cached; N silent peers with instant failures (each probed twice, then no serving peer); five old binaries under the 90 s cap (three probes, then legacy); a slow-`Busy` peer under the cap (three probes, `StillBusy` at 90 s); all Busy (3 backoffs, then `StillBusy`); rejection removes, Busy is re-asked after backoff; Fetch Busy backoff ends exactly at the window end (12 probes); delayed 4.9 s Busy replies (8 probes, 64.9 s); RPC failures retry 5×; non-Busy errors pass through once; watchdog window accounting (2 tests) + 1 in `consensus-types` pinning the BCS tag
- [x] Third commit, same style (paused time, `ForwardEpochSyncService` tested directly with injected build futures): cache hit needs no cold-build permit; concurrent Prepares for one epoch share one build; one cold build per peer at a time; Fetch pool separate from the cold-build pool; LRU keeps the 4 most recently *used* epochs; slow build → `Busy` at 8 s, then served from cache; re-ask after early `Busy` joins the running build (build count stays 1); the reply deadline counts from receipt (a request queued 5 s answers `Busy` after 3 s; an expired deadline still starts/joins the build and answers at once); Fetch lookup never waits; a failed build reaches every waiter and frees the slot. Client: a timed-out peer is re-probed after 5 s (real `RpcError::TimedOut` shape), an unreachable one without a pause, one pause per pass; old-binary and slow-`Busy` deadline tests re-derived for 10 s probes (nine probes / 90 s)
- [x] Fourth commit: one peer flooding Fetch (10 admission attempts while the first is held) holds one slot, another peer is still admitted, the slot is back on drop; same for Prepare, and Prepare/Fetch admissions are independent; the Fetch pool bounds all peers together and dropping one admission frees exactly that slot; Fetch pool separate from the cold-build pool; jitter stays inside its band and reaches both sides of it; the timing-pinned tests pass an exact backoff
- [x] `RUSTFLAGS="--cfg tokio_unstable" cargo test --manifest-path aptos-core/consensus/Cargo.toml --lib forward_epoch_sync` — 35 passed (+ 1 in `consensus-types`)
- [x] Mutation check: dropping the 5 s re-probe pause and skipping the LRU install after a build each turn the corresponding tests red; fourth commit: dropping the per-peer rule (Prepare or Fetch) and dropping the release-on-drop each turn the corresponding tests red
- [x] `cargo clippy -p aptos-consensus -p aptos-consensus-types --all-targets --all-features -- -D warnings`
- [x] `rustfmt +nightly-2026-09-05` on touched files
- [ ] the page loop (`WindowElapsed` → watchdog → `Resume`/continue, `record` per page) has no unit seam (`BlockRetriever` holds a concrete `NetworkSender`); covered by a cluster re-run of the measured epoch: total time unchanged (≈ 27 min), zero `Resume` lines with an honest serving peer


Lchangliang (changes requested):
Requesting changes on `846c421c`.

I confirmed both existing inline findings. In particular, the `BusyExhausted` path has a stronger consensus-liveness impact than the recovery-mode example alone suggests: a Byzantine validator can send a syntactically valid future-epoch message, which reaches `process_different_epoch` before signature verification and makes that sender the preferred retrieval peer. If it keeps replying `Busy`, the client waits 55.5 seconds and then returns an error without ever trying the healthy candidates still in the list. Because `RoundManager` awaits this inline and leaves `wait_change_epoch_flag` unset on failure, the peer can repeatedly retrigger the path and keep the victim from processing proposals, votes, and timeouts. Please continue to the next candidate when one peer exhausts its Busy allowance, and only return the retryable error after all candidates have been considered.

The sequential mixed-version probe also needs a call-level bound. Old peers silently drop the unknown request, so probing every candidate with a 30-second timeout (plus retrying the final peer) changes the worst case from two attempts to roughly `(N + 1) * 30s` before legacy fallback. Please use an absolute Prepare deadline and/or bounded concurrent probes.

There is one additional issue: the advertised 60-second Busy budget is not a wall-clock budget. `BusyBackoff` subtracts only the scheduled sleeps, not time spent in RPCs. Eleven replies arriving just before the 30-second Prepare timeout can occupy the RoundManager for about 385.5 seconds. Fetch creates a fresh `BusyBackoff` for every page, so a serving peer can repeatedly spend the full allowance and then return a small valid batch, with no epoch-level bound. Please carry one absolute deadline across RPC time, sleeps, and all Fetch pages, and make the operation cancellable during RoundManager shutdown.

Please add async tests for: preferred peer always Busy with a healthy alternate; multiple old/unreachable peers under a total deadline; and delayed Busy replies across multiple Fetch pages.

The BCS change itself looks compatible: `Disabled` is appended as tag 7, and I found no persistence/schema or rollback-format issue.

nekomoto911:
Thanks — all three findings confirmed against the code, and the second commit reworks the client around them. One deliberate difference from the requested shape for the Fetch phase, explained below with the measurement that motivated it.

**What changed (commit 2)**

1. *Busy no longer pins one peer.* Prepare runs in passes over the whole candidate list: a `Busy` peer stays for the next pass, an explicit rejection removes the peer, a silent peer gets exactly one more probe (a cold index build past the network layer's 10 s inbound cap is discarded server-side but the build is cached, so the second probe is cheap). A healthy alternate is therefore reached in the first pass without any sleep. Passes that saw `Busy` sleep 0.5/1/2 s, at most 3 times.
2. *One Prepare attempt is capped at 90 s wall-clock* (three 30 s probes; each probe's timeout is cut to whatever is left). At the cap: `Busy` seen → `Resume` (next trigger asks again, never legacy), nothing but rejections/silence → legacy. Concrete bounds: preferred Busy + healthy alternate ≈ the alternate's Prepare RTT, zero sleep; N old binaries → legacy after ≤ 90 s regardless of N (main today: 60 s; first cut: `(N+1)×30 s`); a peer answering `Busy` at 29.9 s → 3 probes, `Resume` at 90 s (first cut: 385 s).
3. *Fetch has an epoch-level bound, as a progress watchdog rather than a wall-clock deadline.* Every 60 s window must deliver ≥ 600 blocks (10 blocks/s) from the serving peer, judged at page boundaries; `Busy` backoff never sleeps past the window end. A failed window returns the new `EpochSyncOutcome::Resume`: everything fetched so far is persisted and replayed, `wait_change_epoch_flag` stays unset, and the next trigger re-Prepares from the advanced `ordered_root`, usually through another peer because triggers rotate across sender keys (cache hit if the same server is re-selected, otherwise one cold build — paid only when a peer actually stalled). `Resume` is logged at info and does not bump `ERROR_COUNT`. This closes the "small valid batch after spending the allowance" case: a 1-block-per-page peer at 4.99 s/page now lasts one window instead of 38 h; a Fetch peer answering `Busy` at 4.99 s per request is dropped after ~65 s.
4. *Cancellable on shutdown.* The `EpochChange` arm races the sync against `close_rx` (`biased`), acks and breaks. `close_rx` can only fire after the node's own pipeline committed the epoch-ending block; every forward page's DB writes complete before the replay await, the legacy path (also covered by the select) writes its blocks and ledger infos before any await, and the next epoch rebuilds the block store from storage, so dropping the future is safe.
5. Tests: the retry loops are free functions over an RPC closure under `tokio::test(start_paused = true)` (tokio `test-util` joins the dev-dependencies): preferred always Busy + healthy alternate (zero sleep); preferred Busy + alternate silent then cached; N silent peers (each probed twice); five old binaries under the 90 s cap (three probes, then legacy); a slow-`Busy` peer under the cap; all Busy (3 backoffs then `StillBusy`); Fetch backoff ending exactly at the window (12 probes); delayed 4.9 s Busy replies (8 probes, 64.9 s); RPC failures 5×; non-Busy errors pass through once; watchdog accounting. The page loop itself (`WindowElapsed` → watchdog → `Resume`) has no unit seam because `BlockRetriever` holds a concrete `NetworkSender`; it is covered by the cluster re-run in the test plan.

**Why a progress watchdog instead of one absolute deadline over the whole sync**

We measured a full-epoch forward sync on a test cluster: a ~30.8k-block epoch takes **26.8 min** (19.2 blocks/s) vs 33.8 min for legacy; legacy finishes *fetching* in ~1.5 min and spends the remaining 32 min in per-block recovery execution, so the sync is execution-bound and forward is only ~26 % faster. A wall-clock deadline short enough to matter against a Byzantine serving peer (≈2 min) would chop every honest sync into ~14 attempts, and each boundary re-selects the serving peer from the trigger's sender, i.e. a cold index build on whichever server has not seen this epoch yet (5–7 s each, or a full 30 s probe timeout when the build exceeds the 10 s inbound cap). A deadline long enough to avoid that (≥10 min) no longer bounds a slow-Busy Fetch peer meaningfully. The watchdog gives the same bound on a bad peer (one window) at zero cost to an honest sync as long as the local replay sustains 10 blocks/s (2× margin over the measured rate; a slower node would `Resume` every minute while still making progress). Prepare is milliseconds for an honest peer, so the absolute cap there is free, which is why it is a plain deadline.

On the "RoundManager cannot process proposals/votes/timeouts" motivation: a node that genuinely needs the epoch sync has no consensus work to do in its current epoch, so the time bound matters for (a) not getting stuck on a bad serving peer and (b) the fake-trigger case (a peer claims a higher epoch before signature verification while the victim is still active). (a) is what the watchdog and the Prepare cap bound. (b) is a #819-era admission-control gap that no per-attempt bound fixes — one message re-triggers the path for free — and is better handled separately; issue filed.

Also filed separately: cold index builds > 10 s never reach the client because `INBOUND_RPC_TIMEOUT_MS` is a hard-coded 10 s in the network layer (the first Prepare always times out and only the cached retry succeeds), and the server-side permit / index-cache hardening.

Lchangliang:
Re-reviewed current head `a2a89150`. I still see three P1 issues that should be resolved before merge:

1. **A `Busy` peer can permanently veto a usable legacy fallback.** `prepare_from_candidates` retains `Busy` peers and returns `StillBusy` after the per-attempt budget; this becomes `Resume`, and the next trigger rebuilds the same candidate set with no cross-attempt Busy count, quarantine, or eventual fallback. With one peer that continuously answers `Busy` and one old peer that times out on Forward Prepare but can serve legacy BlockRetrieval, every attempt returns `Resume` and legacy is never invoked. A bounded total Busy budget must eventually remove/quarantine that peer and allow legacy; long-lived unsupported server states should return `Disabled`, not `Busy`.

2. **Fetch `Resume` can discard all verified cursor progress.** Each attempt initializes its anchor from `ordered_root`, while the watchdog counts raw records and returns `Resume` when the window is below 600. A valid batch may contain no ledger infos; in that case the blocks and QCs are persisted but `recover_blocks_checked` cannot advance `ordered_root` past the old latest LI. The verified `fetch_anchor_*` is only local state and is lost on `Resume`, so the next attempt downloads the same range again. A slow peer can repeat this forever. Persist or reconstruct the verified tail cursor across `Resume`, and measure watchdog progress by monotonic durable cursor/root advancement rather than records received.

3. **Forward RPC admission is not bounded per peer across the full handler.** The dispatcher spawns every request; same-epoch Prepare waiters can join an in-flight build without consuming cold-build or per-peer quota, and Fetch has only a global permit pool. One authenticated peer can occupy all default Fetch permits with concurrent valid requests and keep honest Fetch clients on `Busy`. The Fetch permit is also released before BCS encoding and response delivery, so the quota does not cover the full resource cost. Add a node-wide handler/waiter cap plus per-peer `in-flight = 1`, and hold the guard through response encoding and send. The underlying 100-inbound-RPC-per-connection limit mitigates the waiter count but does not provide the required application-level or node-wide fairness.

Requested tests: persistent Busy plus a legacy-only peer must eventually enter legacy; two Fetch attempts with valid no-LI pages must advance rather than refetch the same range; and one peer flooding concurrent Prepare/Fetch requests must not consume all service capacity.

nekomoto911:
Point 3 confirmed against the code; the fourth commit (`ba4fa7c3`) adds per-peer admission for both request kinds, held through encoding and hand-off, plus jittered `Busy` backoff on the client. Points 1 and 2 get a separate reply.

**What changed (commit 4)**

1. *Per-peer in-flight = 1 for Prepare and for Fetch.* `ForwardEpochSyncService` keeps one set of peers per request kind, keyed on the network-authenticated sender; a peer already in the set is refused with `Busy` (log reasons `prepare_per_peer` / `fetch_per_peer`). Separate sets rather than one because the Fetch set's size is what enforces the node-wide Fetch cap (below), and because a Prepare waiter (no DB work, ≤ 8 s) should neither take nor be blocked by a Fetch slot; an honest client never has both outstanding, so for it the two choices are the same. A peer flooding same-epoch Prepare — the join path, which took no cold-build permit — now holds one handler instead of one per request; a peer flooding Fetch holds one of the four slots and the other three stay open.
2. *Node-wide Fetch cap, from the same set.* With one handler per peer, the number of admitted Fetch peers *is* the number of Fetch handlers running, so `FORWARD_EPOCH_SYNC_FETCH_QUOTA` (default 4, reason `fetch_quota`) caps the set's size in the same critical section and the Fetch semaphore is gone.
3. *The admission covers the whole handler.* It is taken in `process_forward_epoch_sync` before any work and dropped after BCS encoding and `response_sender.send`, so it bounds concurrent handlers *and* the encoding of full-size pages. Before, the Fetch permit was released when `fetch_forward_epoch_sync` returned, so encoding ran outside the quota and a peer could hold a new permit before its previous page was even encoded. The composition with the join path is by construction: the Prepare arm takes the admission before calling into the index lookup, so a joiner holds its peer's single handler for the whole wait. To be precise about "send": the oneshot hand-off to the network layer is the last point the consensus side can observe; from there the network layer copies the page once and holds it in the connection's write queues until the writer task sends it, the same budget every RPC response on that connection shares. What a slow reader leaves buffered there is not this quota's.
4. *`Busy` backoff is jittered (client).* Each `Busy` backoff is drawn uniformly from ±25 % of its nominal length (0.5 → 8 s), for Fetch and for the Prepare passes that saw `Busy`, still clipped to the progress window / attempt deadline as before. A saturated pool refuses everyone at once; with fixed delays those clients come back at the same instant and collide again. The backoff is a parameter of the two retry loops: production passes the jittered one, the timing-pinned tests pass an exact one (`BusyBackoff::exact`, `cfg(test)`).
5. Tests: one peer flooding Fetch (10 admission attempts while the first is still held) holds exactly one slot, another peer is still admitted, and the slot is back once the guard drops; same for Prepare, and Prepare/Fetch admissions are independent; the Fetch pool bounds all peers together (quota 2, three peers; dropping one admission frees exactly that slot); the Fetch pool stays separate from the cold-build pool; the jitter stays inside its band and reaches both sides of it. The pinned-timing tests are unchanged apart from passing the exact backoff. Mutation check: dropping the per-peer rule (Prepare or Fetch) and dropping the release-on-drop each turn the corresponding tests red. The *placement* of the guard (end of `process_forward_epoch_sync`) is by inspection, not by test: the release happens synchronously right after the oneshot send with no yield in between, so no test could observe the difference.

Honest client: no change in the steady state — it has one Prepare outstanding per server and fetches one page at a time. Two corners can now see `Busy` replies where they previously started a second handler. (a) A Fetch retry (after the 5 s RPC timeout, or 500 ms after an RPC error) while the previous page is still being served: it is answered `Busy`, backs off, and is admitted once that handler is gone — the same page is no longer built twice concurrently. (b) A Prepare the client abandoned mid-build (a probe truncated by the 90 s attempt deadline, a cancelled sync, a dropped connection) keeps its slot on the server for at most 8 s from receipt; a re-Prepare inside that window is answered `Busy`, the pass moves on to the other candidates without sleeping, and if that server is the only candidate the attempt can end in `Resume` once before the next one is admitted — no later than the join would have answered, and never legacy.

**Why not a per-peer queue or request merging**

Considered and not done. A per-peer queue would only order a *misbehaving* peer's surplus — an honest client never has a second request outstanding — and without merging, the client's `Busy` backoff is roughly the same wait as a server-side queue, minus the held inbound RPC. Merging same-key Fetches has no honest case at the measured page cost (about 1–2 ms on small pages, estimated up to a few hundred ms on heavy ones, well inside the 5 s client timeout); a page that takes over 5 s means a degraded server where the merged retry would time out again anyway. Slow client-side replay (measured p50 ≈ 0.6 s per batch cycle on light blocks; estimated seconds or more per batch at 5k txn/block) only lengthens the gap between pages, it never produces a second concurrent Fetch.

**What is and is not bounded now**

- Fetch: one handler per peer, `FORWARD_EPOCH_SYNC_FETCH_QUOTA` handlers node-wide, held through encoding and hand-off.
- Prepare: one handler per peer; cold builds `FORWARD_EPOCH_SYNC_COLD_BUILD_QUOTA` node-wide with one initiator per peer (unchanged); every handler is bounded in time by `reply_by` (8 s from receipt). There is no numeric node-wide cap on Prepare *handlers* in this commit. A Prepare handler does no DB work: a waiter holds a task and a shared receiver for at most 8 s, a cache hit encodes one `LedgerInfoWithSignatures`; the only expensive Prepare work, the cold build, is already capped node-wide and per peer. With one handler per peer the count is at most the connected peers — the validator set on the validator network; up to `max_inbound_connections` (default 100) unknown peers per listener without mutual authentication — and the honest population that Prepares simultaneously (every downstream fullnode at each epoch change, plus validators lagging at the boundary) is exactly what a tight number would refuse. If you want a hard bound regardless, a generous one (e.g. 64, one more `len() >= quota` check in `admit_prepare`) is a small addition and I can add it in this PR.
- Identities are free on every listener without mutual authentication: the public network and the VFN network, including a validator's own VFN listener (`mutual_authentication: false` in the validator template), up to 100 unknown inbound connections each. Per-peer = 1 answers the single-peer scenario you raised; it does not give node-wide fairness against `≥ fetch_quota` colluding identities that ignore `Busy`: a freed slot is a race, not a queue, and such a group can keep re-taking it — and the cold-build pool (4 node-wide, 1 per peer) has the same shape for four uncached epochs. Jitter is for honest clients refused together, not for that. The fix there is a bounded FIFO in front of both pools, or the Prepare control plane (cache lookup + join/start decision) as a single serial executor with only the cold build spawned out, which yields a node-wide waiter count for free; I have added it to #836 as a follow-up since it is not covered by that issue as filed.
- The dispatcher still spawns a task per request before admission; that is deliberate — something has to answer `Busy`, and a refused request runs no DB work and takes no lock beyond the admission check.
- This closes #836's first item (per-peer in-flight = 1) for both request kinds; item 3 (trigger admission control) stays out of scope as stated in the description.

Lchangliang (approved):

